PR #173, opened 2013-12-20
Rails 4: Fixed 'Uncaught SyntaxError: Unexpected token ILLEGAL'

semenyukdmitry:
Change related to the way Rails 4 encodes JSON. Long story short: now Rails doesn't escape unicode entities.
http://omegadelta.net/2013/04/28/changes-to-how-rails-3-2-13-and-4-0-encodes-unicode-in-json/


johnmario91:
Still any solution on this? I have the same issue on a Rails 4.0.13 project

PikachuEXE:
I only support 3.x
This is clearly for 2.x

I can merge a PR but I can't fix 2.x myself.

PR #218, opened 2014-06-10
Add "namespace" option to i18n-js.yml

johnnyshields:
In some cases I'd like to export the I18n keys to a different Javascript namespace than `I18n`, to avoid conflicts. For example, I have a widget-type component which is embedded in third-party websites, and many of these sites already define the `I18n` namespace.

To do this, I have added an optional `:namespace` key to `i18n-js.yml`, which can be set per file.

Example:

```yaml
translations:
  - file: "tmp/i18n-js/without_namespace.js"
    only: "*"
  - file: "tmp/i18n-js/with_namespace.js"
    namespace: "MyNamespace"
    only: "*"
```

This generates:

```js
// tmp/i18n-js/with_namespace.js

MyNamespace.translations || (MyNamespace.translations = {});
MyNamespace.translations["en"] = {"test":"Test"};
MyNamespace.translations["fr"] = {"test":"Test2"};
```

In theory this should be a small change to implement, but the `js.rb` (exporter) code is written in a tightly-coupled fashion. I've refactored the code to extract a `Segment` class, which is extensible in case we'd like to add other similar options in the future. All tests are passing, and I've added additional tests specific to the new `Segment` class.

Note that I have not touched the Javascript here, only the Ruby-based exporter code. (JS could be a separate PR)

### DIFF
--- a/lib/i18n/js.rb
+++ b/lib/i18n/js.rb
@@ -4,6 +4,7 @@ require "fileutils"
 module I18n
   module JS
     require "i18n/js/dependencies"
+    require "i18n/js/segment"
     if JS::Dependencies.rails?
       require "i18n/js/middleware"
       require "i18n/js/engine"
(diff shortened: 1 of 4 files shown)


johnnyshields:
Closed in favor of #296

PR #220, opened 2014-06-11
added 'guess' option for missing translation

uris-dev:
Added option missingBehavior, which avoids to have to create strings when developing or simple strings that do not need a declaration in the default language.

- move options comments at declaring array, as this is the right place
- recursive declaration of reset and initializeOptions
- added 'missingBehavior' option with possible values 'message' (or anything else) keeping the actual behavior, 'guess' tries to create a readable string

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 ## unreleased
 
+- You can now set `I18n.missingBehavior='guess'` to have the scope string output as text instead of of the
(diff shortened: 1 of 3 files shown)


rasjani:
https://github.com/rasjani/i18n-js/tree/patch

Same patch on top of current master -  with slight modification to support prefixing the guessed translations.. i'll test before submitting  PR about this ..

rasjani:
this can be closed, same features added in pr #304

Issue #232, opened 2014-06-26
Is there something like I18n.default_scope?


phillipoertel:
This is a question or feature request rather than a bug:

We store all translations for JS in the scope de.client.js. Is it possible to configure the JS for I18n so that it automatically looks for all translations in that scope, without having to supply it with all I18n.t calls? That would be very useful for us.

Example:

// Ruby
I18n.t('de.client.js.foo') # => bar

// JS
I18n.default_scope = "de.client.js";
I18n.t('foo'); # => bar

Thanks and kind regards,
Phillip

lesliev:
I did this slightly differently - I added a helper that I can put into my views:
```ruby
    # Define a Javascript I18n.lt function that allows us to set a prefix in a view, eg:
    # = js_local_translate('admin.assessors.assessment_rounds.components')
    #
    # Then we can use much shorter keys eg:
    # I18n.lt('assessment_round.update')
    #
    def js_local_translate(prefix)
      "<script>
          I18n.lt = function(scope, options = {}) {
            return I18n.t(scope, Object.assign(options, {scope: '#{prefix}'}));
          }
       </script>".html_safe
    end
```

The advantage is that it doesn't interfere with using `I18n.t` for full paths. 
In any case, it would be nice if i18n-js supported something like Rails lazy-lookup: http://guides.rubyonrails.org/i18n.html#lazy-lookup

PikachuEXE:
"lazy-lookup" is based on template path, but there is no template path in JS

PR #256, opened 2014-09-28
Add defaultScope to global config options

dziemian007:
As requested [here](https://github.com/fnando/i18n-js/issues/232) it will add defaultScope global setting that would allow skipping providing the same scope value over and over again. So instead of doing 

```
I18n.t("name", {scope: "activerecord.attributes.user"});
I18n.t("email", {scope: "activerecord.attributes.user"});
```

or

```
var defaults = {scope: "activerecord.attributes.user"};
I18n.t("name", defaults);
I18n.t("email", defaults);
```

we could simply:

```
I18n.defaultScope = "activerecord.attributes.user";
I18n.t("name");
I18n.t("email");
```

What do you think?

### DIFF
--- a/README.md
+++ b/README.md
@@ -420,6 +420,11 @@ You can also provide an array as scope.
     // use the greetings.hello scope
     I18n.t(["greetings", "hello"]);
 
+Default scope can also be configured globally.
+
+    I18n.defaultScope = "activerecord.attributes";
+    I18n.t("user.name");
+
 #### Number formatting
 
 Similar to Rails helpers, you have localized number and currency formatting.
--- a/app/assets/javascripts/i18n.js
+++ b/app/assets/javascripts/i18n.js
@@ -100,6 +100,8 @@
     // string is actually missing for testing purposes, you can prefix the
     // guessed string by setting the value here. By default, no prefix!
     , missingTranslationPrefix: ''
+    // Set default scope for all I18n keys.
+    , defaultScope: null
   };
 
   I18n.reset = function() {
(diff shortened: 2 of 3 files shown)


johnnyshields:
@dziemian007 what do you think about making this a "with_scope" function, such as:

```
I18n.with_scope("activerecord.attributes.user", function(){
  I18n.t("name");
  I18n.t("email");
})
```

My concern here is that your proposal uses the global namespace, and will lead to strange effects if you forget to reset the scope.

PikachuEXE:
Now it really depends on the use case
If one wants the scope permanently, this PR is fine
But if OP wants one-time only scope, then this PR requires something like `withScope`

P.S. @johnnyshields You are in JS land now, naming convention changed :P

johnnyshields:
corrected comment above: with_scope --> withScope

johnnyshields:
I do not see a use case for a global "defaultScope". I think this is bad practice and will likely cause bugs--withScope is much safer IMHO.

PikachuEXE:
```
I18n.defaultScope = "javascript";
I18n.t("save.typing.javascript");
```

dziemian007:
@johnnyshields well I totally agree it could give you a headache when not used properly. I was thinking of ``defaultScope`` more like one-time thing, where you know prefix for all I18n keys used in JS and shorten further usage as much as possible :smile:

johnnyshields:
yeah in practice I still don't see the use case for this. Because your going to have some keys like ".date.formats.long" etc which don't fall under your default scope.

PikachuEXE:
Well I am OK with one-time only approach, since it's safer and no one really asks for the permanent one yet.
@dziemian007 Change the PR to have `I18n.withScope` instead?

johnnyshields:
The more I think about this, the more I think it's not needed.

Why not just do:

```
var myScope = "my_scope.foo.bar."

I18n.t(myScope + "my_key")
I18n.t(myScope + "my_other_key")
```

PikachuEXE:
You better ask guys in #232

johnnyshields:
I still think this needs to be `I18n.withScope(scope, function)` and ideally made thread-safe. Using a global var here will lead to strange behaviors even on single threaded apps.

PR #266, opened 2014-10-27
Add support for loading via AMD and CommonJS

Sija:
Would be super nice to have it baked in! :)


PikachuEXE:
@Sija 
Off topic
Can you give a try on #446 ?
I fixed the case for browser global but not sure if I break AMD/CommonJS case

PikachuEXE:
I am gonna assume it's working fine

Issue #286, opened 2014-11-28
Negative currency does not format correctly


MaryBartlett:
If I feed toCurrency a negative number (-300) then the value returned is $-300; it would be good if the location of the negative symbol could be configured (as I expect it to be -$300).

PikachuEXE:
#290 should solve this
Reopen if issue still exists

PR #290, opened 2015-01-05
Allow forcing currency number sign to be at first place

dziemian007:
As description says, f.ex when doing ``I18n.toCurrency(-1000)`` we get ``$-1,000.00`` in result. This issue was raised already there[https://github.com/fnando/i18n-js/issues/286]. Also ruby I18n does it as we would expect, rather than as it's implemented in i18n-js now:
```
number_to_currency(-9800.56, format: '%u%n')
-€9.800,56 
```
I've changed '-' sign to be at first place by default in this PR, which also can be configured using ``sign_first`` in options. What do you think guys?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@
   Set to `function() {return "";}` to match Ruby `I18n.t("name: %{name}", name: nil)`
 - For date formatting, you can now also add placeholders to the date format, see README for detail
 - Add fallbacks option to `i18n-js.yml`, defaults to `true`
+- Force currency number sign to be at first place using `sign_first` option, default to `true`
 
 ### bug fixes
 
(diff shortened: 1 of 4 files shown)


PikachuEXE:
The only thing is missing is CHANGELOG entry :)
Just do what ruby `I18n` does

PikachuEXE:
Please ping me again if you have made the change, new commits won't notify me :(

dziemian007:
hey @PikachuEXE, thanks for such quick feedback! CHANGELOG updated :)

PikachuEXE:
I see some trailing spaces stripped.
While it's a good practice for coding, doing so in a Markdown document affects the layout.
Fix that and I can merge :)

dziemian007:
Ahh, sorry for that. Fixed now :smile:

PikachuEXE:
Thanks :tada:

PR #293, opened 2015-01-16
Fix link to doc in the config template

stgeneral:
Not sure about the change, maybe the title in the README should be changed.


PikachuEXE:
Use `https://github.com/fnando/i18n-js/blob/master/README.md` or even just `https://github.com/fnando/i18n-js`.
Since it already mentioned the section name (if it does not exist then people will search for it)

PikachuEXE:
Fixed in https://github.com/fnando/i18n-js/commit/80675ba8b7b12143a4c196fd86c6338af258c14a

PR #295, opened 2015-01-28
[WIP] feat(engines): add Engine support

### LINKED ISSUE #262: How to use from a engine
I want to use `i18n-js` inside of a engine (for example `activeadmin`, but could be each other rails engine), now I see one problems:
1. How can I configure `i18n-js` from inside of a engine? The readme shows a [export configuration](https://github.com/fnando/i18n-js#export-configuration), but they only works for inside the app and not from inside a engine!

raszi:
Fixes #262 

I am not sure this is the way you wanted to implement so please feel free to comment I haven't modified the tests yet so they're failing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,9 +110,9 @@ translations:
 ```
 
 #### Export Configuration (For other things)
-- `I18n::JS.config_file_path`  
-  Expected Type: `String`  
-  Default: `config/i18n-js.yml`  
+- `I18n::JS.config_file_paths`
+  Expected Type: `Array`
+  Default: Engine and application roots 'config/i18n-js.yml'
   Behaviour: Try to read the config file from that location  
 - `I18n::JS.export_i18n_js_dir_path`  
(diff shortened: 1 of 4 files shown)


PikachuEXE:
I do plan to move config related methods into some namespace, maybe we should do that first.
And I am not familiarize with engine o_0

Actually I plan on moving `i18n-js.yml` content into ruby form as well! (but that's another topic, I think the gem need a namespace for config related methods)

johnnyshields:
@raszi This looks pretty good, just one thought--what about making a "ConfigSet" object which provides a single interface across all your configs? Currently in the `lib/i18n/js.rb` you're iterating through each config individually which makes for lots of extra code.

PikachuEXE:
I actually opened #298 
Currently blocked by a few PR
#318 is almost done
#325 was just submitted and requires some review

Then I will start working on this

PikachuEXE:
Please check #326

PR #296, opened 2015-01-31
Config options: Namespace, PrettyPrint, and Export I18n JS

johnnyshields:
This commit adds the following features:
- Add `:namespace` option to config (per segment) which sets the Javascript namespace in the translations output file in order to avoid namespace conflicts
- Add `:pretty_print` option to config (per segment) which writes the output file with linebreaks and indentation
- Add `:export_i18n_js` to config (global) which functions the same as `I18n::JS.export_i18n_js_dir_path`

In addition, I have refactored the following:
- Extract "Segment" class which handles writing of translation JSON files, and allows options to be set per-segment. Includes specs.
- Move some specs to new `sprockets_spec.rb` and `utils_spec.rb` files to reduce bloat
- Minor change of `I18n::JS.export_i18n_js_dir_path` so that any non-`String` is interpreted as "do not export". This should not break existing implementations, and also the README has previously said the only way you can set path is via a `String` variable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,20 +110,33 @@ translations:
 ```
 
 #### Export Configuration (For other things)
-- `I18n::JS.config_file_path`  
+
+- `I18n::JS.config_file_path`
   Expected Type: `String`  
   Default: `config/i18n-js.yml`  
   Behaviour: Try to read the config file from that location  
-- `I18n::JS.export_i18n_js_dir_path`  
+
+- `I18n::JS.export_i18n_js_dir_path`
   Expected Type: `String`  
   Default: `public/javascripts`  
   Behaviour:  
   - Any `String`: considered as a relative path for a folder to `Rails.root` and export `i18n.js` to that folder for `rake i18n:js:export`
-  - `nil`: Disable `i18n.js` exporting
+  - Any non-`String` (`nil`, `false`, `:none`, etc): Disable `i18n.js` exporting
+
+- You may also set `export_i18n_js` in your config file, e.g.:
+
+```yaml
+export_i18n_js_: false
+# OR
+export_i18n_js: "my/path"
+
+translations:
+  - ...
+``
 
 To find more examples on how to use the configuration file please refer to the tests.
 
-##### Fallbacks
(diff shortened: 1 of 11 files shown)


johnnyshields:
@PikachuEXE please consider merging this soon.

ovikholt:
:+1:

PikachuEXE:
First, I think this pull request too too big to be reviewed properly.
Second, I am very busy for my job until mid-March :s
Third, I actually plan to move the config out of yml. (#298)

bjornstar:
Having pretty print would be really helpful!

johnnyshields:
@PikachuEXE

1) Most of the PR code is just tests for the new functionality. All I've done is re-factored the currently bulky export process to be slightly more object oriented, i.e. introducing a Segment object which encapsulates the JSON generation and can take options.

2) Can anyone else help review?

3) Can you give an example of what you mean by moving config out of yml? My design in this PR is that each export file should take its own generation options--this is very useful for my app. Example:

```
translations:
- file: "public/javascripts/file_1.js"
  pretty_print: true
  only:
    - '*.date.abbr_day_names'
    - '*.date.month_names'
    - '*.date.abbr_month_names'
  

- file: "some/other/dir/file_2.js"
  namespace: "CustomI18n"  
  only:
    - 'en.my.key'
```

PikachuEXE:
You can ignore point 3 for now.
Point 1 is related to point 2.
The diff is quite big so I can't review it with the time I have.

johnnyshields:
@PikachuEXE would it help for me to break into 3 PRs:
- Segment change (refactor)
- `namespace` + `pretty_print` options
- `export_i18n_js` option

PikachuEXE:
@johnnyshields yessss
I feel painful when I have to deal with many "big" internal PR in my company :sos:

johnnyshields:
OK done. Raised #299, #300, #301, and #302.

Please merge 299-301 in order.

302 can be merged by itself

PikachuEXE:
Will check soon! Thanks!

PR #299, opened 2015-02-05
Extract Segment class from lib/i18n/js.rb

johnnyshields:
Extract Segment class from lib/i18n/js.rb. This models an single "output file" which contains a set of translations for one or more locales.

This structure is extensible to add options in the future.

### DIFF
--- a/spec/segment_spec.rb
+++ b/spec/segment_spec.rb
@@ -1,0 +1,35 @@
+require "spec_helper"
+
+describe I18n::JS::Segment do
+
+  let(:file)        { "tmp/i18n-js/segment.js" }
+  let(:translations){ { "en" => { "test" => "Test" }, "fr" => { "test" => "Test2" } } }
+  let(:options)     { {} }
+  subject { I18n::JS::Segment.new(file, translations, options) }
+
+  describe ".new" do
+
+    it "should persist the file path variable" do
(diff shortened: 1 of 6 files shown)


spec/segment_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+describe I18n::JS::Segment do
+
+  let(:file)        { "tmp/i18n-js/segment.js" }
+  let(:translations){ { "en" => { "test" => "Test" }, "fr" => { "test" => "Test2" } } }
+  let(:options)     { {} }
+  subject { I18n::JS::Segment.new(file, translations, options) }
+
+  describe ".new" do
+
+    it "should persist the file path variable" do
```
PikachuEXE: Shouldn't we use `its(:file)` directly?
`.should` is old syntax.
Just saying, no need to change it now.

johnnyshields: we should use either:

```
its(:file) { is_expected.to eq("tmp/i18n-js/segment.js") }

# OR

expect(subject.file).to eq("tmp/i18n-js/segment.js")
```

PikachuEXE: I rather use the first one, otherwise setting `subject` is kind of pointless.
And it also reads better!

PR #300, opened 2015-02-05
Config options: Namespace and PrettyPrint

johnnyshields:
This commit adds the following features:
- Add `:namespace` option to config (per segment) which sets the Javascript namespace in the translations output file in order to avoid namespace conflicts
- Add `:pretty_print` option to config (per segment) which writes the output file with linebreaks and indentation

Please merge #299 first

PR #301, opened 2015-02-05
Export i18n js config

johnnyshields:
- Add `:export_i18n_js` to config (global) which functions the same as `I18n::JS.export_i18n_js_dir_path`
- Minor change of `I18n::JS.export_i18n_js_dir_path` so that any non-`String` is interpreted as "do not export". This should not break existing implementations, and also the README has previously said the only way you can set path is via a `String` variable.

Please merge PR #299 and #300 first

### DIFF
--- a/README.md
+++ b/README.md
@@ -182,6 +182,36 @@ You must disable this feature by setting the option to `false`.
 To find more examples on how to use the configuration file please refer to the tests.
 
 
+#### Namespace
+
+Setting the `namespace` option will change the namespace of the output Javascript file to something other than `I18n`.
+This can be useful in no-conflict scenarios. Example:
+
+```yaml
+translations:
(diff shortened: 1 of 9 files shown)


johnnyshields:
Here's a diff of master vs. this branch: fnando:master...johnnyshields:export-i18n-js-config

PikachuEXE:
Sorry don't have time to review this today :S

PikachuEXE:
Is there a reason to set it to `:none` instead of `nil` when it's not a string?
Just asking

johnnyshields:
Because `nil` evaluates as `false` (unset) when using the `||=` operator. By using the symbol `:none` it means the user has explicitly said "I do not want to export this file" somewhere along the chain. This was the easiest way to preserve the existing behavior of this function.

PR #302, opened 2015-02-05
Cleanup: split large test file into several smaller files

johnnyshields:
This commit simply moves test code from one place to another. No functional change.


PikachuEXE:
Sorry forgot this one :P

PR #304, opened 2015-02-18
reworked 'guess' missing behaviour patch.. 

rasjani:
Added option missingBehaviour, which avoids to have to create strings when developing or simple strings that do not need a declaration in the default language.
Combine that with missingTranslationPrefix and you can still identify which strings are translated and which are guessed.

This commit is based on pr#220 with some small changes added..

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,11 @@
 
 ## Unreleased
 
+- You can now set `I18n.missingBehavior='guess'` to have the scope string output as text instead of of the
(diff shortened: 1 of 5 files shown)


PikachuEXE:
Related to #220

PikachuEXE:
Sorry can you you rebase on master?
I think my change on CHANGELOG causes conflict, sorry :P

rasjani:
pr updated against upstream/master

PikachuEXE:
Thanks! I will release another version soon :)

PR #307, opened 2015-02-23
2.X - Updated url used to retrieve js library from Github.

bendangelo:
In the old 2.X branch the js file was being pulled from the wrong location; I fixed it.


PikachuEXE:
Thanks for your PR!
But I believe they should now use `https://raw.githubusercontent.com/fnando/i18n-js/2.x/vendor/assets/javascripts/i18n.js`, if they are using 2.x
Care to make the change?

bendangelo:
Fixed

PR #311, opened 2015-03-03
Added specs for requirejs amd loading

thegillis:
I wasn't aware there was already a fix in master for loading i18n.js file over requirejs. Since the fix is already applied, here is the test I used to verify the bug: spec/js/specs_requirejs.html

It only requires a small change to the translations javascript file to detect amd loading.

### DIFF
--- a/spec/js/require.js
+++ b/spec/js/require.js
@@ -1,0 +1,2083 @@
+/** vim: et:ts=4:sw=4:sts=4
+ * @license RequireJS 2.1.16 Copyright (c) 2010-2015, The Dojo Foundation All Rights Reserved.
+ * Available via the MIT or new BSD license.
+ * see: http://github.com/jrburke/requirejs for details
+ */
+//Not using strict: uneven strict support in browsers, #392, and causes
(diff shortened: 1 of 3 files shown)


PikachuEXE:
Can this also be auto run?
Just curious, not a blocker.

PikachuEXE:
The spec seems to fail since `I18n` is always undefined

thegillis:
Alright, I clearly have a commit problem. Let me double check what I have and see if I can get those tests to pass.

thegillis:
Alright, should all pass now.

I wrote the tests not knowing the missing "return" bug in rc8 was already fixed in master. My fix did not explicitly name the module 'i18n' in the define call:

https://github.com/fnando/i18n-js/blob/master/app/assets/javascripts/i18n.js#L21

I added a mapping to fix the module loading.

I have not yet looked at how auto running the js specs work.

PikachuEXE:
All seems fine now :)

PikachuEXE:
I am going to study how to run it from command line

PR #312, opened 2015-03-05
Add except option to config

### LINKED ISSUE #125: Exclude only certain keys
Hi!

I'm working on a Rails App that uses https://github.com/gregbell/active_admin. Since it will clutter the I18n namespace with `activeadmin.*` keys, is there any way to exclude them (the opposite `only` on i`18n-js.yml`)?

I've tried to fiddle a little bit with the code but I can't seem to find a simple way to do it.

Thanks!

pjg:
It allows you to exclude certain keys from both the exported JS translations file and from the file generated for the Rails' assests pipeline.

Example:

```yaml
  translations:
    - except: ['active_admin', 'ransack']
```

This will make it so that the generated/exported JS file including all translations will skip any phrases which are listed under the `active_admin` or `ransack` keys (which makes sense as you don't need translations from those gems in the frontend; they only unnecessarily make your translations file bigger).

This configuration is valid for each "segment" in the config file, so you can have various exception rules for each segment.

PS: We needed that option to slim down our exported JS file (it has become too big).

Closes #125


PikachuEXE:
Will let people review for this weekend and merge next week.
Then release another version

pjg:
Awesome! Looking forward for any feedback :)

PikachuEXE:
If no one object to this change (including option name, how it works, etc.) today.
I will merge this tomorrow and release a new version.

PikachuEXE:
Since no one objects to this PR, merging :)

PikachuEXE:
Will release a new version today or tomorrow
Got an issue #313 not sure if need fixing

pjg:
Thank you!

Issue #314, opened 2015-03-11
RequireJS optimization doesn't correctly detect define method in i18n.js


thegillis:
I'm currently incorporating the i18n-js gem into my project. We currently export translations and use the requirejs-rails gem to load the i18n.js file from the server. We then will use the r.js optimization tool to assemble all of our dependencies into one JS file.

Although the contents of the i18n.js file are included in this r.js optimized version, it does not correctly detect the define method and appends a "shim" define at the end. Unfortunately this causes the I18n JS object to be inaccessible. I have tracked the problem down to the global closure inside the define call.

Steps to reproduce (current version of r.js included in the latest requirejs-rails gem on ubuntu server with packages nodejs npm):

```
wget 'https://raw.githubusercontent.com/jwhitley/requirejs-rails/0.9.5/bin/r.js'
# master copy as of 3/11/2015
wget 'https://raw.githubusercontent.com/fnando/i18n-js/f67fa51427dd004b43f20c46c4016a71012d6699/app/assets/javascripts/i18n.js'
# Optimized version
nodejs r.js -o baseUrl=. name=i18n out=i18n-optim.js
# Non-optimized version
nodejs r.js -o baseUrl=. name=i18n out=i18n-nooptim.js optimize=none
```

If you view the i18n-nooptim.js file, the last line is:

```
define("i18n.js", function(){});
```

Unfortunately this causes all require statements to return undefined for the 'i18n' dependency:

```
require(['i18n'], function(I18n) {
  I18n.t('true'); // FAIL
});
```

This is also identical to the most recent development copy of the r.js tool:

```
# master as of 3/11/2015
wget -O r.2.1.16.js 'https://raw.githubusercontent.com/jrburke/r.js/443809bfe5d668bf4217c4f26b4636011cf91aa6/dist/r.js'
nodejs r.2.1.16.js -o baseUrl=. name=i18n out=i18n-16-optim.js
nodejs r.2.1.16.js -o baseUrl=. name=i18n out=i18n-16-nooptim.js optimize=none
```

You will notice that the i18n-16-nooptim.js and i18n-16-optim.js files suffer from the same problem.

The fix is to replace the line:

```
define('i18n', (function(global){ return function(){ return factory(global); }})(this));
```

With:

```
var global=this;
define('i18n', function(){ return factory(global);});
```

I believe that these are functionally identical, but I'm not sure about the original intention for this define statement statement, so I'm not sure if it will violate any edge cases. It's even fewer bytes :)

[...]

PR #315, opened 2015-03-11
RequireJS optimization doesn't correctly detect define method in i18n.js

thegillis:
Example fix for bug #314.

### DIFF
--- a/app/assets/javascripts/i18n.js
+++ b/app/assets/javascripts/i18n.js
@@ -18,7 +18,8 @@
     module.exports = factory(this);
   } else if (typeof define === 'function' && define.amd) {
     // AMD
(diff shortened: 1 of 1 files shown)


PikachuEXE:
Can you run test added in #311?
I can't get auto test for require JS working yet :S

thegillis:
Both the spec.html file and spec_requirejs.html file run without errors:

spec.html:
i18n_spec

spec_requirejs.html
i18n_requirejs

PikachuEXE:
Really need some automatic test for this kind of stuff Q_Q

PikachuEXE:
@thegillis 
Off topic
Can you give a try on #446 ?
I fixed the case for browser global but not sure if I break AMD/CommonJS case

PikachuEXE:
I am gonna assume it's working fine

Issue #239, opened 2014-07-15
translations.js is being generated differently between different machines


rafaelgaspar:
The order of the keys are different.

This can't happen with Asset Pipeline, otherwise my application.js is going to be different across my app servers, and that causes a different asset fingerprint.

I'm trying to look further into this, but still haven't figured after what commit this started happening, will update if I find something.

PikachuEXE:
Someone has kindly submitted a PR: #318
So I guess now would be great to discuss and decide what we want:
- sorted translation to ensure consistency for translation files generated on different machines
- keep translation unsorted, resolve by chaning gems like [`russian`](https://github.com/yaroslav/russian/) 
- **both** (with option), default unsorted

Envek:
Second case (just keep unsorted) is not an option as it's hard to patch all the gems (especially not actively maintained ones, see yaroslav/russian#56, there are still no feedback). And I saw a lot of neglected gems recently.
I think current behavior (unsorted) with option enabling sorting will satisfy everyone. Good explanation in README is required too.
Personally, I do not care for order of keys at all.

@msjonker, can you please tell you opinion? And thanks for pull request!

msjonker:
The only reliable solution we found to ensure consistent asset fingerprints was to sort prior to JSON. I agree that this could be turned on as an option, for those who don't want this. However, I don't think the sorting of the JSON should have any adverse effects, aside from the extra processing.

PikachuEXE:
I guess we can add it as an option, but warn about this "unsorted" behaviour until next major version
And then always sort it for next major version (if no one objects)
That option can be default to true (since they can set it back and we are still in `rc`)

msjonker:
That works for me.

PikachuEXE:
#318 is almost done.
Could you guys review it (or even try it) to see if that's what you want.
Once reviewed I will merge.

Envek:
Looks good for me, but can't battle test it now.

PikachuEXE:
I will wait until you have battle test it :clock3:

Envek:
Well, I don't sure that I will be able to battle test it all — we are no more working on the machines where I've noticed this.

PikachuEXE:
@rafaelgaspar Are you able to test #318?

rafaelgaspar:
Sorry, I ended up having to scope my JS translations better to avoid this bug, so I can't reproduce it anymore in my app.

PikachuEXE:
Well then I will just merge it.

[...]

PR #318, opened 2015-03-24
Deep sort hash keys before converting to JSON

msjonker:
Fixes the issue in #239 by deep sorting the translation keys before converting to JSON output


PikachuEXE:
Let's continue the discussion in #239 to decide this PR is suitable solution or not,
thank you for your PR first :)

PikachuEXE:
As discussed in #239, can you add an option to change the behaviour?
(Defaults to `true`, and just add the option to `I18n::JS`, I will move options to another place later)

I have no idea how do the "warning" thing though (when it's set to `false`).

msjonker:
Yeah, I'll try and put something together in the next week.

PikachuEXE:
@msjonker Do you have time to look at this yet?

msjonker:
I added I18n::JS.sort_translation_keys to control the sorting of translation keys. It is true by default. I did not add a warning, however I did add some documentation to the README. I don't know that a warning should be necessary and I, too, could not decide on a way to implement that.

spec/i18n_js_spec.rb
```diff
@@ -422,4 +422,80 @@
...
+      context 'when :sort_translation_keys set to true in config' do
+        before { set_config "js_sort_translation_keys_true.yml"}
+
+        it { should eq true }
+      end
+
+      context 'when :sort_translation_keys set to false in config' do
+        before { set_config "js_sort_translation_keys_false.yml"}
+
+        it { should eq true }
```
PikachuEXE: When ":sort_translation_keys set to false in config" it should equal `true`???

spec/i18n_js_spec.rb
```diff
@@ -422,4 +422,80 @@
...
+      end
+    end
+
+    describe '.sort_translation_keys?' do
+      after { described_class.send(:remove_instance_variable, :@sort_translation_keys) }
```
PikachuEXE: Why not use `before`?

spec/i18n_js_spec.rb
```diff
@@ -422,4 +422,80 @@
...
+      context "when it is set to false" do
+        before { described_class.sort_translation_keys = false }
+
+        it { should eq false }
+      end
+    end
+
+    context "sort_translation_keys option" do
+
+      subject {
```
PikachuEXE: Use `do...end`

spec/utils_spec.rb
```diff
@@ -66,4 +66,23 @@
...
+      result.should eql({:y => 3, :z => {:a => 2, :b => 1}})
+    end
+
+    it "performs a deep keys sort without changing the original hash" do
+      hash = {:z => {:b => 1, :a => 2}, :y => 3}
+
+      result = described_class.deep_key_sort(hash)
+
+      result.should eql({:y => 3, :z => {:a => 2, :b => 1}})
+      hash.should eql({:z => {:b => 1, :a => 2}, :y => 3})
```
PikachuEXE: Create a new variable which clones `hash`, avoiding the chance of typo

PikachuEXE:
@msjonker Do you have time to look at my line notes? :)
I can cherry pick & fix them myself if you don't have time

spec/i18n_js_spec.rb
msjonker: This needs to be performed after, similar to the describe '.export_i18n_js_dir_path' tests, in order to reset the instance variable so that it doesn't persist to the next test. This is why the incorrect false test pointed out above passed incorrectly.

msjonker:
Code review items have been fixed.

PikachuEXE:
Last step: add CHANGELOG entry
Please add under `Unreleased` instead of the last released version ;)
Had a few PR did that

I will let people in #239 have a look at this a few days and then I will merge next week

lib/i18n/js.rb
```diff
@@ -71,11 +71,12 @@ def self.merge_with_fallbacks!(result)
     end
 
     def self.filtered_translations
-      {}.tap do |result|
+      translations = {}.tap do |result|
         translation_segments.each do |segment|
           Utils.deep_merge!(result, segment.translations)
         end
       end
+      Utils.deep_key_sort(translations) if I18n::JS.sort_translation_keys?
```
PikachuEXE: This makes `filtered_translations` to return `nil` when `I18n::JS.sort_translation_keys?` is `false`

PikachuEXE: Add an example to ensure this doesn't happen after fix

PikachuEXE:
I will fix the remaining problems :)
Thanks for your work!

PikachuEXE:
Regression fixed in b21d5b373467212e7d9c9b205c487c22fc8e350e

Issue #298, opened 2015-02-05
Move config out of yaml file, into some nameplace like `I18n::JS::Configuration`


PikachuEXE:
I understand we can just use ERB inside the yaml file.
But the config options are not very nested options, and there are many more new proposed options which only reside in root namespace.
The only nested option now seems to be `translations`, but still not very deep nested.

Also there are some config options written in ruby are in the module root namespace, which seem confusing to me. I suggest moving them into a namespace.

PikachuEXE:
Going to start working on this one

Issue #20, opened 2010-09-15
Rails 3 support


fnando:
Include Railtie for Rails 3.

PR #319, opened 2015-03-28
Extract methods from I18n::JS to new Exporter module, and clarify public interface

johnnyshields:
Currently the I18n::JS library has a lot of code in the base file `lib/i18n/js.rb`. This is an anti-pattern and makes the code difficult to work with.

As a first step in refactoring, I've moved much of the internals into a new class `I18n::JS::Exporter`, and I've exposed only the key methods in `I18n::JS`. In subsequent PRs I hope to break `Exporter` into even smaller units.

The big chunks of code from `lib/i18n/js.rb` and `spec/i18n-js.rb` have been moved to respective `exporter` lib/spec files.


PikachuEXE:
Thanks for submitting!
I wanted to do it in the next release anyway :D
But I am not sure if the config should be placed in the `Exporter` or not

I can see 2 patterns:
1. Place all config in `Configuration` class, let people `I18n::JS.configure do |config|`
2. Place all config in different related modules, like `config_file_path`

I can see many gems using pattern 1, but I don't remember seeing pattern 2 elsewhere (except in this PR)

Any thoughts?

PikachuEXE:
Maybe wait for #318 & #20 to be merged first
And then remove the changes for config, it can be a separate thing to handle

johnnyshields:
We can do both a `Configuration` class AND this `Exporter` change. The two are not mutually exclusive.

The configuration should be able to be set via Ruby `I18n::JS.configure do |config|` AND also loaded from `config/i18n-js.yml`; as long as there's only one set of config internal variables in-memory we can be flexible about where it is loaded from.

johnnyshields:
To clarify: I agree that `Config` does NOT belong in `Exporter`. I only put the config in the `Exporter` class to minimize the amount of code change.

PikachuEXE:
Since I am not crystal clear about how to change the config related thing now, making this PR to only extract exporting related methods would make it easier to be merged.

We should discuss config related thing in #298

johnnyshields:
@PikachuEXE I think it would be easiest to refactor both the config and the exporter in the same PR, or else do the config first.

PikachuEXE:
Then let's wait for me to come up a PR for config extraction first.
Still that needs to wait for #318 & #320 to be merged first.

johnnyshields:
@PikachuEXE I think we can merge this in advance of the config extraction. The two PRs do separate things--this one concerns where the exporting logic is located (regardless of config). I will rebase this tonight so that it can be merged.

johnnyshields:
Actual, no rush on this. Lets do config PR first.

PR #320, opened 2015-03-28
Unify/simplify the behavior of per-locale and non-per-locale files by moving per-locale generation to Segment class

johnnyshields:
This PR unifies/simplifies the behavior of per-locale and non-per-locale files.

Specifically, splitting files per-locale generation is done in the `Segment` class `.save!` method, which is the final step before writing the file. (Previously, "per-locale" and "non-per-locale" files followed very different logic paths which diverged early in the processing chain.)

One positive effect of this change is that the "fallbacks" feature now works for both per-locale and non-per-locale files.

**Important:** There are two breaking changes in this PR, hence I recommend a minor version bump.

1. Per-locale files should now have `only`/`exclude` scopes which begin with '*', which is the same as non-per-locale files.

2. Since this PR fixes the issue that fallbacks were ignored for non-per-locale files, users may see some behavior change. for this. This is unavoidable as it's essentially a bug that this wasn't done before.

Lastly, please note that I had to add `fallbacks: false` to many of the test fixtures to ensure behavior consistency.

### DIFF
--- a/lib/i18n/js.rb
+++ b/lib/i18n/js.rb
@@ -107,9 +107,9 @@ module I18n
 
       [scopes].flatten.each do |scope|
         translations_without_exceptions = exclude(translations, exceptions)
-        filtered_translations = filter(translations_without_exceptions, scope)
+        filtered_translations = filter(translations_without_exceptions, scope) || {}
 
-        Utils.deep_merge! result, filtered_translations || {}
+        Utils.deep_merge!(result, filtered_translations)
       end
 
       result
--- a/lib/i18n/js.rb
+++ b/lib/i18n/js.rb
@@ -34,17 +34,6 @@ module I18n
       translation_segments.each(&:save!)
     end
 
-    def self.segments_per_locale(pattern, scope, exceptions, options)
-      I18n.available_locales.each_with_object([]) do |locale, segments|
-        scope = [scope] unless scope.respond_to?(:each)
-        result = scoped_translations(scope.collect{|s| "#{locale}.#{s}"}, exceptions)
-        merge_with_fallbacks!(result, locale, scope, exceptions) if use_fallbacks?
-
-        next if result.empty?
-        segments << Segment.new(::I18n.interpolate(pattern, {:locale => locale}), result, options)
-      end
-    end
-
     def self.segment_for_scope(scope, exceptions)
       if scope == "*"
         exclude(translations, exceptions)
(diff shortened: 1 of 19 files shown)


spec/i18n_js_spec.rb
```diff
@@ -263,6 +283,8 @@
   end
 
   context "I18n.available_locales" do
+    # before { allow(I18n::JS).to receive(:fallbacks).and_return(false) }
```
PikachuEXE: Commented code o_0

lib/i18n/js/segment.rb
```diff
@@ -14,12 +16,21 @@ def initialize(file, translations, options = {})
...
+          I18n.available_locales.each do |locale|
+            write_file(file_for_locale(locale), self.translations.slice(locale))
+          end
+        else
+          write_file(self.file, self.translations)
+        end
+      end
 
-        File.open(self.file, "w+") do |f|
+      def write_file(_file, _translations)
```
PikachuEXE: Either make it a class method
Or make it `def write_file(_file = self.file, _translations = self.translations)` (I am not sure if this works)
It seems pointless to have an instance method 
that always takes some arguments that sometimes can be retrieved in the object itself

lib/i18n/js/segment.rb
```diff
@@ -14,12 +16,21 @@ def initialize(file, translations, options = {})
 
       # Saves JSON file containing translations
       def save!
-        FileUtils.mkdir_p File.dirname(self.file)
+        if file =~ LOCALE_INTERPOLATOR
+          I18n.available_locales.each do |locale|
+            write_file(file_for_locale(locale), self.translations.slice(locale))
```
PikachuEXE: Either always use `self.something`, or don't use `self` all the time (except using setter)
code with mixed style is hard to read

lib/i18n/js.rb
```diff
@@ -111,7 +109,7 @@ def self.scoped_translations(scopes, exceptions = []) # :nodoc:
         translations_without_exceptions = exclude(translations, exceptions)
         filtered_translations = filter(translations_without_exceptions, scope)
 
-        Utils.deep_merge! result, filtered_translations
+        Utils.deep_merge! result, filtered_translations || {}
```
PikachuEXE: Can you add brackets?

johnnyshields:
All comments have been addressed.

PikachuEXE:
For your breaking changes:
- For (2), it's fine, it was a bug
- For (1), It's not clear to me for what you are talking about.
  Please clarify it (or write README and add CHANGELOG entry directly :P

johnnyshields:
Old format. Note that the %{locale} one doesn't begin with '*'

```yaml
- file: '/my_dir/not_per_locale.js'
  only:
    - '*.foo.*'
    - '*.bar.*'

- file: '/my_dir/my_%{locale}.js'
  only:
    - 'foo.*'
    - 'bar.*'
```

New format:

```yaml
- file: '/my_dir/not_per_locale.js'   # same
  only:
    - '*.foo.*'
    - '*.bar.*'

- file: '/my_dir/my_%{locale}.js'   # now has *
  only:
    - '*.foo.*'
    - '*.bar.*'
```

PikachuEXE:
OK, I think that's more consistent and sensible.
Please update README and CHANGELOG so I can merge this.
Thanks :D

johnnyshields:
Updated README and CHANGELOG

PikachuEXE:
Common mistake: you have added the changelog entry for a released version o_0

johnnyshields:
Fixed

michieldewit:
For the `except` option the `*.something.*` format is not supported as is for the `only` option. It only supports key names. So when `errors` is specified for the `except` option, all of the following segments will be excluded: `errors`, `app.errors`, `app.view.errors`. It seems to have always been like this. Please adjust comments in CHANGELOG/README or change the functionality (the latter I would prefer).

PikachuEXE:
@michieldewit Please open another issue :)

michieldewit:
I created a pull request for this: #325

PikachuEXE:
Will review it next week <3

PR #321, opened 2015-04-07
JS file: set count to 1 by default

johnnyshields:
I think it's useful to set count to 1 by default. The user can override it, but it saves the user from having to set count in lots of places when you use a lot of pluralizations.

Currently if you have a pluralized key and you DON'T set count 1, you see `[Object object]`. There's also no harm setting count: 1 on a non-pluralized key.

### DIFF
--- a/app/assets/javascripts/i18n.js
+++ b/app/assets/javascripts/i18n.js
@@ -331,7 +331,7 @@
   //
   I18n.prepareOptions = function() {
     var args = slice.call(arguments)
-      , options = {}
+      , options = {count: 1}
       , subject
     ;
 
@@ -344,10 +344,6 @@
 
       for (var attr in subject) {
         if (!subject.hasOwnProperty(attr)) {
-          continue;
-        }
-
-        if (this.isSet(options[attr])) {
           continue;
         }
 


PikachuEXE:
I don't think it's a great approach to sliently "rescue" such usage.
I believe that most of the time when option is not set, they actually forgot to set it.

johnnyshields:
As someone who does a lot of I18n (10 languages in my app), I really think this would help.

At a minimum, we need to make this say "missing translation". But I really do think it's better to set count to 1 all the time, in practice it makes life very easy. Please reopen this issue.

PikachuEXE:
So there are 2 issues here:
- It's not saying "missing option" when option required is missing
- You want to be able to apply "default option" when calling `I18n.translate` (where the default options you is `{count: 1}`

I suggest making this PR to solve the 2nd issue.
You can add a new option to set "default options" when calling `I18n.translate`, which would:
- fulfill your need
- keep the default behaviour "safe" (well it doesn't return proper error message yet, but it's another issue)
- other developers can use it for other "default options"

PikachuEXE:
Also if you want to reuse existing changes, beware that `I18n.prepareOptions` is `The precedence is from left to right` (written in comments).
See the failed build

johnnyshields:
OK, I think defaultOptions makes sense here. I will raise a separate PR for that.

PR #325, opened 2015-05-08
Absolute exception specifiers with wildcard support

michieldewit:
Implemented absolute (i.e. scoped) `except` filters. The filters for `only` are always absolute (i.e. relative to the root of the translations structure), whereas `except` filters were always relative (i.e. matching local key names only).

This change adds support for absolute `except` filters, so that specific branches from the translations structure can be excluded by specifying an absolute path.

Examples:
- `except: "*.errors"` would exclude all translations in the `error` root scope for all languages, but would leave a scope like `en.devise.errors` in tact.
- `except: "*.countries.*.official"` would leave out the official country names (e.g. `en.countries.us.offical`), but would not exclude informal names (e.g. `en.countries.us.informal`).

This changes gives much more control over exclusions, preventing accidental exclusions.

To provide backwards compatibility, all unqualified `except` filters (i.e. WITHOUT periods) are considered relative and evaluated using the old logic, whereas all qualified filters (i.e. WITH period) will be evaluated as absolute. Please note that previously, `except` filters with a period were always ignored.

### DIFF
--- a/lib/i18n/js.rb
+++ b/lib/i18n/js.rb
@@ -124,7 +124,7 @@ module I18n
       exceptions.inject(translations) do |memo, exception|
         exception_scopes = exception.to_s.split(".")
         Utils.deep_reject(memo) do |key, value, scopes|
-          key.to_s == exception.to_s || Utils.scopes_match?(scopes, exception_scopes)
+          Utils.scopes_match?(scopes, exception_scopes)
         end
       end
     end
--- a/lib/i18n/js.rb
+++ b/lib/i18n/js.rb
@@ -122,8 +122,9 @@ module I18n
       return translations if exceptions.empty?
 
       exceptions.inject(translations) do |memo, exception|
-        Utils.deep_reject(memo) do |key, value|
-          key.to_s == exception.to_s
+        exception_scopes = exception.to_s.split(".")
+        Utils.deep_reject(memo) do |key, value, scopes|
+          key.to_s == exception.to_s or Utils.scopes_match?(scopes, exception_scopes)
         end
       end
     end
--- a/lib/i18n/js.rb
+++ b/lib/i18n/js.rb
@@ -124,7 +124,7 @@ module I18n
       exceptions.inject(translations) do |memo, exception|
         exception_scopes = exception.to_s.split(".")
         Utils.deep_reject(memo) do |key, value, scopes|
-          key.to_s == exception.to_s or Utils.scopes_match?(scopes, exception_scopes)
+          key.to_s == exception.to_s || Utils.scopes_match?(scopes, exception_scopes)
         end
       end
     end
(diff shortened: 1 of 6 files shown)


michieldewit:
Btw, this PR is the result of a discussion in #320, where it was assumed that `except` filters would allow wildcards. Since `except` actually did not allow wildcards (because these filters are always relative), I decided to implement absolute `except` filters with wildcards.

johnnyshields:
+1 looks good to me.

My only comment is that we need to move all this hash mangling logic its own class, but that can be done in a separate PR to minimize change.

PikachuEXE:
Oops got some merge conflict since I merged #318 
Please rebase :)

michieldewit:
Rebased it

lib/i18n/js.rb
```diff
@@ -122,8 +122,9 @@ def self.exclude(translations, exceptions)
       return translations if exceptions.empty?
 
       exceptions.inject(translations) do |memo, exception|
-        Utils.deep_reject(memo) do |key, value|
-          key.to_s == exception.to_s
+        exception_scopes = exception.to_s.split(".")
+        Utils.deep_reject(memo) do |key, value, scopes|
+          key.to_s == exception.to_s or Utils.scopes_match?(scopes, exception_scopes)
```
PikachuEXE: Please use `||` instead of `or`

lib/i18n/js/utils.rb
```diff
@@ -22,10 +22,18 @@ def self.deep_merge!(target_hash, hash) # :nodoc:
...
+          unless block.call(k, v, scopes + [k.to_s])
+            memo[k] = v.kind_of?(Hash) ? deep_reject(v, scopes + [k.to_s], &block) : v
+          end
+        end
+      end
+
+      def self.scopes_match?(scopes1, scopes2)
+        if scopes1.length == scopes2.length
+          [scopes1, scopes2].transpose.all? do |scope1, scope2|
+            scope1.to_s == '*' or scope2.to_s == '*' or scope1.to_s == scope2.to_s
```
PikachuEXE: Please use `||` instead of `or`

PikachuEXE:
Also there is no need for "backward compatibility".
This feature is added in `3.0.0.rc9`, which is NOT considered as stable version 
(well actually RC should be relative stable, but 3.x.rc means beta instead :S, off topic)

PikachuEXE:
Oh please add an example for "old syntax" to ensure it ..fails?
I don't want 2 different behaviours for `only` and `except`

michieldewit:
I removed the 'backwards compatibility' support for relative `except` filters.

PikachuEXE:
LGTM
Please add CHANGELOG entry.
Then I will merge

michieldewit:
@PikachuEXE Added a CHANGELOG entry

PikachuEXE:
Good! :)
Thank you

PR #326, opened 2015-05-13
Refactor/configuration

PikachuEXE:
# Background
- Too many config related accessors attached to `I18n::JS`
- Engine support is requested (#262)
# Breaking changes
- Configuring through YML replaced by pure Ruby way (for now, YML can be re-added later)
- Configuration methods moved (they might be referenced already)
# Why have Ruby config support instead of pure YML file(s)?
- Had [a PR for support of ERB in YML file](https://github.com/fnando/i18n-js/pull/224)  
  Meaning _some_ people wants dynamic config generation  
- Pure ruby way can support "engine" (or other gems trying to config this gem)  
  YML can also do that, but seems a bit complex (See #295)  
- Usage of special ruby objects, like `Regexp`, is easier in Ruby  
  Also if there is a config to support  
- YML is good for static deep nested structure data, like locales  
  The most deep nested key in this gem's config is `translations`  
  Meaning the current config structure is not very deep nested, less need for YML  
# TODO
- [x] Extract configuration into a separate class
- [x] Make the gem configurable with Pure Ruby
- [x] Rename some config options
- [ ] Provide YAML support (for backward compatibility)
- [ ] Wait for response (loop)
- [ ] Update doc
- [ ] Update CHANGELOG
# Config options to be checked for renaming
- [x] `export_i18n_js_dir_path` => `i18n_js_export_path`
- [x] `sort_translation_keys` (unchanged)
- [x] `fallbacks` (unchanged)
- [x] `translations` => `translation_segment_settings`
# Similar PR
- #295 
- #319 
# Related PR
- Fix caching expiration by changing (related) config value (#355)


johnnyshields:
@PikachuEXE I think we should continue to support configuring through YAML. Just my preference. It's a pain to have both an initializer and an i18n-js.yml file.

johnnyshields:
Otherwise I'm fine with the refactors to consolidate the config variables into a class.

lib/i18n/js.rb
```diff
@@ -76,32 +85,26 @@ def self.filtered_translations
...
     def self.translation_segments
-      if config? && config[:translations]
-        configured_segments
-      else
-        [Segment.new("#{DEFAULT_EXPORT_DIR_PATH}/translations.js", translations)]
-      end
-    end
+      configuration.translations.inject([]) do |segments, options|
+        file = options[:file]
+        only = options[:only] || '*'
```
raszi: Maybe:
```ruby
options.fetch(:only, '*')
```

lib/i18n/js/configuration.rb
```diff
@@ -0,0 +1,73 @@
...
+      end
+
+      ## Setting this to nil would disable i18n.js exporting
+      def export_i18n_js_dir_path=(new_path)
+        new_path = :none unless new_path.is_a?(String)
+        @export_i18n_js_dir_path = new_path
+      end
+
+      def sort_translation_keys?
+        @sort_translation_keys = true if @sort_translation_keys.nil?
```
raszi: This method could be simplified to this:

```ruby
@sort_translation_keys ||= true
```

PikachuEXE: Then it will always be `true`, not more `false`

lib/i18n/js.rb
PikachuEXE: I will do more refactor once this is merged
But I don't want to change too many things in one PR

PikachuEXE:
@johnnyshields PR description updated

johnnyshields:
@PikachuEXE I still disagree with this and I think there's very simple way to support most if not all of the config nodes in YML. We could easily add a YamlConfigurer class which overrides the config when the YAML is loaded, maybe move everything to a "config" node.

PikachuEXE:
Maybe can have a "config loader" thing
Then `I18n::JS.configuration = I18n::JS::YamlConfiguration.new("some/path")` ?
(just concept, not suggesting exact code)

johnnyshields:
I think we can place the YAML config to a `config:` node in the root of the YAML file for cleanliness. Anytime a YAML file is read with a config node, it is auto-parsed by the I18n::JS::YamlConfigurer class (which just funnels the values into the main Ruby config variables).

PikachuEXE:
For "engines", they can use the pure Ruby way
Not sure how it works when there is an "engine" and a YAML config file in app...

johnnyshields:
My understanding is that engines just means "multiple `i18n-js.yml` files at different places in the app." So in that case, if they do use the `config:` node twice you just parse it twice and overwrite values. (No special logic needed for engines specifically)

PikachuEXE:
I imagine code of an engine would have some code like
https://github.com/yaroslav/russian/blob/master/lib/russian.rb#L30

They just change the config by Ruby code **before** app config
So for config option like `translations`, the README should have instructions for:
- gem/engine authors: Use `unshift`/`push`
- developers: Use `unshift`/`push`
(I am not sure if the order of hashes affects the output)

PikachuEXE:
YAML might be added after this, but it won't be included in this PR.
This PR is to make things works like it's a new gem.

johnnyshields:
@PikachuEXE YAML is supported currently, I do not think we should break this in a PR.

(You're welcome to make a PR that drops YAML support as long as we flag it "work in progress" and do not actually merge it)

PikachuEXE:
TODO updated
Need some feedback on config option naming first
Naming is one of the most difficult tasks for me

johnnyshields:
`export_i18n_js_dir_path`  -->  `export_i18n_js_path`   (or `_dir`) is better I think.  Others are fine IMHO

PikachuEXE:
`export_i18n_js_path`? 
`i18n_js_export_path`?

I think `translations` reads a bit... ambiguous
But don't know what to rename to (`translation_somethings`)

johnnyshields:
`i18n_js_export_path` is the best.

`translations` is fine because it's the main config option.

PikachuEXE:
Will go with `i18n_js_export_path`
I will come up some names for `translations` later within this week

PikachuEXE:
I suggest `translations` => `translation_segments`.
Name taken from existing methods [`I18n::JS.translation_segments`](https://github.com/fnando/i18n-js/blob/refactor/configuration/lib/i18n/js.rb#L92)

Speaking of that method, I wonder where should the logic for filtering translations be moved to.
But let's discuss that in #333

johnnyshields:
Let's keep the name as `translations` for now. It's very intuitive; `translation_segments` is not.

johnnyshields:
about filtering logic, my PR here https://github.com/fnando/i18n-js/pull/319 was a first step to cleaning up all this logic--I think the first thing to do is to move it out of the `I18n::JS` module and into a new temporary module called `I18n::JS::Exporter`. We can then break the `Exporter` into even smaller classes.

PikachuEXE:
1. OK I will keep the old name.
2. Discuss filtering in #333

PikachuEXE:
Too busy interviewing people for this week and maybe next week also.
Pausing until those stuff done

PikachuEXE:
Final got a chance to look at the YAML config support
I actually need to rewrite again since too many conflicts when rebasing

I have only renamed the options on ruby side, the option names in YAML file are still the same.
Also `translations` renamed to `translation_segment_settings`

PikachuEXE:
@johnnyshields @raszi 
Any comment?
I hope this can be merged within 2 weeks since it was stalled too long (because of me :S

johnnyshields:
@PikachuEXE I will try this with my app over the weekend and let you know. Thanks for your good work and for YAML support.

PikachuEXE:
Thanks for trying it out!
Also try to Ruby config once if you can :)

README.md
```diff
@@ -74,93 +74,89 @@ Then get the JS files following the instructions below.
...
+      file: 'public/javascripts/path-to-your-messages-file.js',
+      only: '*.date.formats',
+    },
+    {
+      file: 'public/javascripts/path-to-your-second-file.js',
+      only: ['*.activerecord', '*.admin.*.title'],
+    },
+  ]
+end
+```
```
johnnyshields: we should keep the YAML examples, or show both Ruby and YAML for each.

README.md
```diff
@@ -724,6 +758,31 @@ I18n.translations["pt-BR"] = {
 }
 ```
 
+## Using YAML file as config store
+Earlier versions of this gem was mainly using YAML config file for loading file.  
+And it was changed to have Ruby object as core configuration implementation.  
+For those using YAML config file and don't want to convert it into an initialzer,
+A loader class is created for loading such file.
+YAML config file will be supported for `3.x`, but might deprecated and removed in next major release.
```
johnnyshields: I don't agree with this. YAML should have first-class support along with Ruby. It shouldn't be "maybe deprecated".

johnnyshields:
This broke by current config.

My config is at `config/i18n-js.yml`. It looks something like this:

```yaml
export_i18n_js: false

translations:
- file: 'my/dir/path/assets/javascripts/table_solution/i18n/translations.js'
  pretty_print: true
  only: &TABLE_SOLUTION
    - '*.time.formats.time'
    - '*.date.formats.default'

- file: 'my/other/path/assets/javascripts/widget/i18n/translations.js'
  namespace: 'MyI18nNamespace'
  only:
    - '*.widget.*'
    - '*.date.month_names'
```

Looks like it's maybe not reading the config at all. Please ensure the `export_i18n_js, pretty_print, namespace` etc option are working.

```

File: README.md [hunk 5]
```diff
@@ -724,6 +758,31 @@ I18n.translations["pt-BR"] = {
...
+  # Change this to be the path of your YAML config file
+  Rails.root.join("config", "i18n-js.yml"),
+).load
+```
+
+### YAML key names to Ruby config attribute name Mappings
+- `sort_translation_keys` => `sort_translation_keys`
+- `fallbacks` => `fallbacks`
+- `translations` => `translation_segment_settings`
+- `export_i18n_js` => `i18n_js_export_path`
```
johnnyshields: These should be made consistent.

README.md
```diff
@@ -724,6 +758,31 @@ I18n.translations["pt-BR"] = {
...
+### Usage
+```ruby
+# Yes you still need some ruby code to load the YAML file,
+# but it should be easier than converting the whole YAML if it's large.
+
+require "i18n-js"
+::I18n::JS::Configuration::YamlFileLoader.new(
+  # Change this to be the path of your YAML config file
+  Rails.root.join("config", "i18n-js.yml"),
+).load
```
johnnyshields: YAML file should be automatically detected, without the need for extra config.

johnnyshields:
OK I got it working. Here are my comments:
* [ ] There's a bug that the YamlFileLoader is not symbolizing keys, and the Segment class isn't able to read. You should change:

          ```rb
          if config_from_file.key?(:translations)
            configuration.translation_segment_settings = config_from_file.fetch(:translations)
          end
          ```

          to something like:

          ```rb
          if config_from_file.key?(:translations)
            configuration.translation_segment_settings = (config_from_file.fetch(:translations) || []).map!(&:symbolize_keys)
          end
          ```

* [ ] Propose to rename `YamlFileLoader` to YamlConfigReader
* [ ] Would like to see Ruby and YAML examples side-by-side in readme.
* [ ] Please remove note in Readme about YAML being deprecated in the future.
* [ ] YAML config should be automatically detected and loaded if present at `config/i18n-js.yml` (as it is now)
* [ ] Would like if Ruby and YAML config key names can be made consistent.
* [ ] For some reason, the Javascript output has changed to `I18n.extend((I18n.translations["en"] || {}), ...`. I'd like to have a config option to disable this; I like it how it was before.

README.md
PikachuEXE: Agree, will change that

README.md
PikachuEXE: I doubt, gem should make more things explicit.
A developer just need to set up a gem one time.
Creating an initializer file for a gem is very common for a rails app.

johnnyshields: It's also very common for gems to detect a `.yml` file in the `/config/` dir, e.g. Mongoid, etc.

PikachuEXE: Mongoid only use YAML file, there is no 2nd choice.
Having an opt-in approach is safer, and encourages user think before using Pure Ruby / YAML file for config.
Also this PR is also raised for #262, so that should also be considered.

johnnyshields: I still think there's no harm in detecting the default file `/config/i18n-js.yml`. If it's not there, then ignore it, and you can still manually include configs using Ruby (e.g. for engines)

PikachuEXE: I can add it afterward.
I want to avoid any automatic thing in this PR to so it safer for #262.
Reviewing more at once == bigger chance to miss something.
e.g. I forgot to append instead of overwriting `translation_segment_settings` when loading YAML file.

johnnyshields: I think we should (re)add it at this time. Otherwise you are breaking the currently supported pattern. Principle of least surprise.

PikachuEXE: I won't release this before re-adding the "auto-detect".
My point is the change set for this PR is already very large.
I am coding as if the YAML config feature does not exist before.

PikachuEXE:
I name the loader class as `YamlFileLoader` since it's loading **one** file and it's already under `Configuration` namespace.

Thanks for spotting the bug in `Segment`, I will add more spec & fix it.

I will make Ruby and YAML examples side-by-side.

YAML config file detection will be handled in another PR after this is merged.

`I18n.extend` is added recently (see #354).
But your request for new option does not belong to this PR.

The `translation_segment_settings` in YAML config will still overwrites the config in Ruby.
I am not sure how to "solve" this.
The YAML
But that request does not belong to this PR.

PikachuEXE:
@raszi @timoschilling 
Can you take a look at the following proposed change and see if it solves #262?

#### How should `translation_segment_settings` in YAML config file change the config option `translation_segment_settings` in Ruby?

### Current: Replace entire content
This is contradicting one of the purposes of this PR: to support engine
If the content is replaced, all config setup by engines will be replaced.

### Solution: Append only
This will allow settings added by engines to retain.
But there won't be default value for translation_segment_settings anymore
since the default value only exists expecting it to be overwritten, not appended to.

PikachuEXE:
Aiming this for 4.0.0

Too many changes are included in this PR
And I am too busy :S

PikachuEXE:
This PR is too old lol
Will revisit but won't be this PR

PR #329, opened 2015-05-25
Fix directory validation

andresgap:


PikachuEXE:
According to http://ruby-doc.org/stdlib-2.2.0/libdoc/fileutils/rdoc/FileUtils.html#method-c-mkdir_p
It does nothing it exists already
What's the motive of this PR?

PikachuEXE:
Add comment above the changed line:
`path could be a symbolic link`

Then add CHANGELOG entry

PikachuEXE:
Please use master for now, got #326 pending before next release
Also feel free to comment on that PR (need more people reviewing it)

Issue #338, opened 2015-06-22
Fallbacks don't work when translations are set to nil


michieldewit:
I am using a tool to manage translations online (WebTranslateIt). This tool makes sure all language keys are available in all YML files. If no translation is available, it will output a `nil` value (`~` in YML). The new `I18n::JS::Utils MERGER` does not take these `nil` values into account and will not merge the appropriate fallback.

Example:

```
en:
  key:
    value: "My value"
```

Spanish (`es`) is configured with `en` as fallback:

```
es:
  key:
    value: ~
```

The resulting JavaScript should be something like:

```
I18n.translations["es"] = {
  "key": {
    "value": "My value"
  }
}
```

but instead is

```
I18n.translations["es"] = {
  "key": {
    "value": null
  }
}
```

PikachuEXE:
Please try `master` to see if this is resolved

michieldewit:
It is. Thanks!

PR #339, opened 2015-06-22
Implemented fix for issue #338

michieldewit:


PikachuEXE:
I just wonder what happen if false is encountered...

michieldewit:
@PikachuEXE `false` is a value and would overwrite the value in the target hash. In casu: when, say, a Spanish translation file contains a `false` value for a certain key, that would overwrite the value for the, say, English fallback. In general `false` is considered a value, whereas `nil` is the absence of a value. So it makes sense not to overwrite a fallback translation with a value that means 'no value'.

PikachuEXE:
"merged" as 24376065a8c0f3c5802bc2e3aa1b444a59729fab

PR #349, opened 2015-07-28
Allow translations.js to be exported to specified directory

joecorcoran:
Previously there was no way of setting this, so it always went
to the default directory, which is not always desirable.

This commit writes translations.js to the same directory as i18n.js
when running the export rake task.

### DIFF
--- a/lib/i18n/js.rb
+++ b/lib/i18n/js.rb
@@ -85,7 +85,7 @@ module I18n
       if config? && config[:translations]
         configured_segments
       else
-        [Segment.new("#{DEFAULT_EXPORT_DIR_PATH}/translations.js", translations)]
+        [Segment.new("#{export_i18n_js_dir_path}/translations.js", translations)]
       end
     end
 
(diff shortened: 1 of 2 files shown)


PikachuEXE:
Thanks for your PR!
I am kinda of busy right now, will look at it on Friday.
I can't trust myself on busy days.

PikachuEXE:
I guess this doesn't work yet
Since `export_i18n_js_dir_path` could be `:none`, indicating to NOT export `i18n.js` when running export rake task.

PikachuEXE:
Also it seems possible to have some config `translation` in `i18n-js.yml` to achieve what you want.
See `Export Configuration (For translations)` in README

afdev82:
I'm trying to use the option to export in 'app/assets/javascripts/i18n' (to use both with the manifest file for js and to load the translations server side to execute with ExecJs).
But this configuration doesn't work: 

```
export_i18n_js: "app/assets/javascripts/i18n/i18n.js"
translations:
  - file: 'app/assets/javascripts/i18n/translations.js'
```

PR #352, opened 2015-08-11
Support :exclude as an alias for :except

johnnyshields:
Rails uses the terminology `:only` and `:exclude`. It was a bit unnatural when I found this was called `:except`


PikachuEXE:
I think I will look at this after config refactor :S
But this is valid

Might as well rename the whole option z_z

johnnyshields:
Why not merge this as-is? I think there's no problem supporting both names.

PikachuEXE:
Having "aliases" creates need for extra testing.
If this is a "rename option" PR, I can accept it before the refactor.

PikachuEXE:
@johnnyshields 
Do you still want this?
If so please update rebase and let me know

PikachuEXE:
Oh please update README too

johnnyshields:
@PikachuEXE please merge

johnnyshields:
You're right, I think it's better to keep only and except. I will close this PR it's not needed

PR #354, opened 2015-08-20
Allow extending of translation files

AntiFish03:
@fnando 
    I added the ability for multiple translation files to be added to a single page.  The reason I felt this was so important was so a global and route specific translation file could be precompiled into the rest of my code base properly.  This could also be used on a widget or module level as well as long as the top level properties don't clobber each other.

### DIFF
--- a/README.md
+++ b/README.md
@@ -586,6 +586,130 @@ The accepted formats are:
 
 Check out `spec/*.spec.js` files for more examples!
 
+## Using multiple exported translation files on a page.
+This method is useful for very large apps where a single contained translations.js file is not desirable. Examples would be a global translations file and a more specific route translation file.
+
+### Rails without asset pipeline
+1. Setup your `config/i18n-js.yml` to have multiple files and try to minimize any overlap.
+  ```yaml
+  sort_translation_keys: true
+  fallbacks: false
+
+  translations:
+    + file: "app/assets/javascript/nls/welcome.js"
+      only:
+        + '*.welcome.*'
+      pretty_print: true
+    
+    + file: "app/assets/javascript/nls/albums.js"
+      only:
+        + '*.albums.*'
+      pretty_print: true
+
+    + file: "app/assets/javascript/nls/global.js"
+      only:
+        + '*'
+      # Exempt any routes specific translations from being
+      # included in the global translation file
+      except:
+        + '*.welcome.*'
+        + '*.albums.*'
+      pretty_print: true
+  ```
+  When `rake i18n:js:export` is executed it will create 3 translations files that can be loaded via the `javascript_include_tag`
+
+2. Add the `javascript_include_tag` to your layout and to any route specific files that will require it.
+  ```ruby
+    # views/layouts/application.html.erb
+    <%= javascript_include_tag(
+          "i18n"
+          "nls/global"
+        ) %>
+  ```
+  and in the route specific
+
+  ```ruby
+    # views/welcome/index.html.erb
+    <%= javascript_include_tag(
+          "nls/welcome"
+        ) %>
+  ```
+
+3. Make sure that you add these files to your `config/application.rb`
+
+  ```ruby
+    config.assets.precompile += %w(
+      i18n
+      nls/*
+    )
+  ```
+
+### Using require.js / r.js
+
+To use this with require.js we are only going to change a few things from above.
+
+1. In your `config/i18n-js.yml` we need to add a better location for the i18n to be exported. You want to use this location so that it can be properly precompiled by r.js.
+
+  ```yaml
+  export_i18n_js: "app/assets/javascript/nls"
+  ```
+
+2. In your `config/require.yml` we need to add a map, shim all the translations, and include them into the appropriate modules
+
+  ```yaml
+  # In your maps add (if you do not have this you will need to add it)
+  map:
+    '*':
+      i18n: 'nls/i18n'
+
+  # In your shims
+  shims:
+    nls/welcome:
+      deps:
+        + i18n
+
+    nls/global:
+      deps:
+        + i18n
+
+  # Finally in your modules 
+  modules:
+    + name: 'application'
+      include:
+        + i18n
+        + 'nls/global'
+    
+    + name: 'welcome'
+      exclude:
+        + application
+      include:
+        + 'nls/welcome'
+  ```
+3. When `rake assets:precompile` is executed it will optimize the translations into the correct modules so they are loaded with their assigned module, and loading them with requirejs is as simple as requiring any other shim.
+  ```javascript
+  define(['welcome/other_asset','nls/welcome'], function (otherAsset){
+
+  });
+  ```
+4. (optional) As an additional configuration we can make a task to be run before the requirejs optimizer. This will allow any automated scripts that run the requirejs optimizer to export the strings before we run r.js
+
+  ```rake
+  # lib/tasks/i18n.rake
+  Rake::Task[:'i18n:js:export'].prerequisites.clear
+  task :'i18n:js:export' => :'i18n:js:before_export'
+  task :'requirejs:precompile:external' => :'i18n:js:export'
+
+  namespace :i18n do
+    namespace :js do
+      task :before_export => :'assets:environment' do
+        I18n.load_path += Dir[Rails.root.join('config', 'locales', '*.{yml,rb}')]
+        I18n.backend.load_translations
+      end
+    end
+  end
+
+  ```
+
 ## Using I18n.js with other languages (Python, PHP, ...)
 
 The JavaScript library is language agnostic; so you can use it with PHP, Python, [your favorite language here].
--- a/app/assets/javascripts/i18n.js
+++ b/app/assets/javascripts/i18n.js
@@ -883,20 +883,14 @@
   I18n.p = I18n.pluralize;
 
   if ( typeof define === "function" && define.amd ) {
-    if (require.specified('i18n')) {
-      window.requirejs.undef('i18n');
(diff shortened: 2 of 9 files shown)


AntiFish03:
Also I have included the changes to allow requirejs to properly define I18n as a module and return the correct object.

PikachuEXE:
I am not sure if this should belong to I18n (JS)

It is basically a utility function that merge 2 objects together
And there are many libraries what includes it already (you can even write your own version)

AntiFish03:
@pikachuEXE,  it's true that there are a lot of libraries that contain this or similar code.  However because of the way the exported files work either you have to force the selection of one in the gem or have a small internal implementation of an extend.

AntiFish03:
Also now that I am not on my phone, I would like to point out that all the exported files use
 ```javascript
I18n.translations[locale] = { ... }
```

This is fine if you are only loading 1 file for a given locale but what if I would like to load a global translation file and a route specific translation file, instead of an en/de/fr....  the above breaks and it gets even worse when you add in something like requirejs/r.js optimizer and I have no control over the order that dependencies load. and given that it uses a window variable to attach to there is not a good way to prevent one translation file from over writing the other.  The solutions as I see them:
1) To do the extend as I have done.
2) Setup the translations to be loaded as JSON and do an ajax from the frontend for them.
3) Only load 1 translation file which given the amount of translations that a large app might have is a problem for load times.

1 translation file is what I am trying to avoid given that my primary js file is already pretty large.  and adding it to the route files is not an option because that goes against a lot of programming best practices(DRY, and most caching strategies also).  I need to be able to properly separate the translation files by common/global scoping and the route specific files.  The PR that I have made represents what I think is the best option for solving the issue. I am sharing it because I doubt I am the only one that has run into this issue. and it is a very simple and elegant solution.

**Comment by PikachuEXE** [comment]
I agree with your use case.

My question is: 
There are many versions of `extend` function in existing libraries or as a code snippet.
Why I18n is responsible of providing **one** version of `extend` that is not related to the project's functionality?

If you are trying to share your approach of separating translation files, you can write an article and post it to a blog or something. You can also submit a PR with adding link of your article to this project's doc.

**Comment by AntiFish03** [comment]
@pikachuEXE and again the issue is the way the export files are generated. If they were just JSON files we wouldn't even be having this conversation.   But given they are javascript files they should not conflict , or conflict as little as possible with other javascript files. Which means there needs to be some kind of functionality to facilitate the process.

**Comment by PikachuEXE** [comment]
I feel like this change requires a section in README describing the use case it's supporting (the global & route specific translation files).
Please add it to README directly and I will review again.

**Comment by AntiFish03** [comment]
I can do that... Its simple enough

File: app/assets/javascripts/i18n.js [hunk 1]
```diff
@@ -857,12 +854,43 @@
...
+   * Merge obj1 with obj2
+   * @param {Object} obj1 Default settings
+   * @param {Object} obj2 User obj2
+   * @returns {Object} Merged values of obj1 and obj2
+   */
+  I18n.extend = function ( obj1, obj2 ) {
+    var extended = {};
+    var prop;
+    for (prop in obj1) {
+      if (Object.prototype.hasOwnProperty.call(obj1, prop)) {
```
PikachuEXE: Just wondering what versions of JS (ES3, ES5, ES6...) supports `Object.prototype` & `hasOwnProperty`

app/assets/javascripts/i18n.js
```diff
@@ -857,12 +854,43 @@
     }
 
     return scope;
-  }
+  };
+  /**
+   * Merge obj1 with obj2
```
PikachuEXE: Maybe put a note to clarify that this is a shallow merge.
Also obj1 or obj2 is not changed

brauliobo:
@AntiFish03 @PikachuEXE I think the README changes are good enough, not so sure if the `require.js` stuff are relevant enough.

app/assets/javascripts/i18n.js
brauliobo: it seems to be from ES3: http://stackoverflow.com/questions/8157700/object-has-no-hasownproperty-method-i-e-its-undefined-ie8

brauliobo: The jquery extend version (https://github.com/jquery/jquery/blob/compat/src/core.js#L119) is more like the simple version from prototype: http://stackoverflow.com/a/929791/670229

PikachuEXE: I see.
Would be great to add the reason of using just code in comments.
Good for learning & tracing code changes

brauliobo: @PikachuEXE I would let this go for now and fix later if needed

PikachuEXE: I will add it myself then (later) :3

AntiFish03: The extend method was meant to be as simple as possible. I used the KISS method for it... you could  copy jQuery, Underscore, etc... extend method if you chose to but they are much more complex and could lead to instability.

brauliobo: @AntiFish03 agree, I think you version is simple enough.

PikachuEXE: Maybe there is some misunderstanding here.
I have read [the page](http://stackoverflow.com/questions/8157700/object-has-no-hasownproperty-method-i-e-its-undefined-ie8) and I agree there should be not issue with this function supporting ES3.

In additional to the code, if there is external sources of evidence supporting the decision of having this version of implementation, let's add them in comment.
Other dev can save some time by just visiting the URL, instead of having to find the line comment in this PR.

Doc + Test > Implementation

spec/js/extend.spec.js
```diff
@@ -0,0 +1,41 @@
+var I18n = require("../../app/assets/javascripts/i18n")
+  , Translations = require("./translations")
+;
+
+describe("Extend", function () {
```
PikachuEXE: Can we add a test case for "NOT changing the original input objects"?
Well since jQuery one does that :S

PikachuEXE:
Since the `I18n.extend` here only does shallow merge,
I think a note about "Don't split translations by second level keys"

### Example
They **can** put
```yaml
translations:
    + file: "app/assets/javascript/nls/welcome.js"
      only:
        + '*.welcome.*'
```

They **cannot** put
```yaml
translations:
    + file: "app/assets/javascript/nls/welcome/messages.js"
      only:
        + '*.welcome.messages.*'
    + file: "app/assets/javascript/nls/welcome/actions.js"
      only:
        + '*.welcome.actions.*'
```

README.md
```diff
@@ -586,6 +586,130 @@ The accepted formats are:
...
+1. Setup your `config/i18n-js.yml` to have multiple files and try to minimize any overlap.
+  ```yaml
+  sort_translation_keys: true
+  fallbacks: false
+
+  translations:
+    + file: "app/assets/javascript/nls/welcome.js"
+      only:
+        + '*.welcome.*'
+      pretty_print: true
```
PikachuEXE: @AntiFish03 
Maybe that's what you are using.
But `pretty_print` doesn't have to be true to have `multiple exported translation files`
Could you remove it to avoid confusion?

brauliobo:
@PikachuEXE LGTM

PikachuEXE:
Will merge after one final review :)
@AntiFish03 I just found that a CHANGELOG entry is missing.

AntiFish03:
@PikachuEXE I apologize that I have not completed this... I am very busy with work... I will attempt to get the change log added and remove the pretty print from the readme.

PikachuEXE:
@AntiFish03
No worry!
I think many people are busy at work usually and I appreciate your contribution with your spare time.
Just ping me again when you are done changing the PR :)

AntiFish03:
@PikachuEXE I added the PR to the change log.

PikachuEXE:
Thanks!
Sorry for taking this long.

PikachuEXE:
@AntiFish03 
[I reverted some code changes](commit_with_reverted_change) (not related to `extend`) that causes error in JS spec
See https://travis-ci.org/fnando/i18n-js/builds/84036584

[commit_with_reverted_change]: [https://github.com/fnando/i18n-js/commit/4178100bb85f03b09bf79558eb984efa250b8c83]

casertap:
I use this in my application.js:
```
//= require_directory ../../../phrase/locales
```
This means all the files generated via: rake i18n:js:export will be included in my application.js
For instance I have those files: phrase.en-AU.js, phrase.en.js phrase.vi.js
At the end I18n.translations will be only equal to the last traduction because there is no merge of the translations json property
```
var I18n = I18n || {};            
I18n.translations = {"en":{"jobs": .....
```
So it get erased on each included phrase.*.js file.

How do I do to have the following behavior?
```
```

jakemhiller:
I think this breaks the use case where the translation files are included before `I18n.js`

https://github.com/fnando/i18n-js#setting-up

`Uncaught TypeError: I18n.extend is not a function`

Maybe that wasn't intended to work in the first place, but it made keeping `I18n.js` in my main js file a lot easier, since the translation file could be included separately above that.

PikachuEXE:
I agree.
I guess what you are describing is a new issue.
Could you open a new issue and link back to this one?
If you know how to make the file compatible so that the main library files and translations can be loaded with any order, feel free to open a PR directly

PikachuEXE:
@AntiFish03 @jakemhiller @brauliobo 
Sorry for bothering you in this old PR
We have #416 which intends to change `I18n.extend` from shallow merge to deep merge.
Please comment on that PR if you got any opinion. 
Thanks :)

PR #371, opened 2015-11-24
Fix of missing initializer at sprockets.

vimutter:
They are using Railtie, so there are no special initializer. I've got 'immutable env' error. Which was fixed by that.

### DIFF
--- a/lib/i18n/js/engine.rb
+++ b/lib/i18n/js/engine.rb
@@ -3,7 +3,7 @@ require "i18n/js"
 module I18n
   module JS
     class Engine < ::Rails::Engine
-      initializer "i18n-js.register_preprocessor", :after => "sprockets.environment" do
+      initializer "i18n-js.register_preprocessor", after: :engines_blank_point, before: :finisher_hook do
         next unless JS::Dependencies.using_asset_pipeline?
         next unless JS::Dependencies.sprockets_supports_register_preprocessor?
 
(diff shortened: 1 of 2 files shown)


PikachuEXE:
Any document that support this change?
I am not an expert for rails initializer and I think most people are not as well.
It will be better for everyone to know what options should be used and why.

lib/i18n/js/engine.rb
```diff
@@ -3,7 +3,7 @@
 module I18n
   module JS
     class Engine < ::Rails::Engine
-      initializer "i18n-js.register_preprocessor", :after => "sprockets.environment" do
+      initializer "i18n-js.register_preprocessor", after: :engines_blank_point, before: :finisher_hook do
```
vimutter: Ok, little bit more details:

1. Here is where sprockets are attached with Rails. There is no `'sprockets.environment'` mentioned.
https://github.com/rails/sprockets-rails/blob/master/lib/sprockets/railtie.rb

2. Finisher hook is the place which should be used as border. 
http://guides.rubyonrails.org/configuring.html#initializers

https://github.com/rails/sprockets-rails/blob/master/lib/sprockets/railtie.rb#L160

vimutter:
@PikachuEXE Added line comment with more details.

PikachuEXE:
I was expecting those to be put in comment, but I can add them afterwards

Also found a solid reference to "sprockets.environment"
https://github.com/rails/sprockets-rails/issues/227

Can you add a change log entry (in `bug fixes`)?

vimutter:
@PikachuEXE Done, please check.

PikachuEXE:
Thanks. Merging.

PR #372, opened 2015-12-01
readme: known issues rewrite

chaimleib:
Use `rake tmp:cache:clear` if your version of Rails is too old and you don't have `assets:clobber`


PikachuEXE:
Thanks for the rewrite :)
Merging

Issue #365, opened 2015-10-28
toNumber loses precision on big numbers


andrewjtait:
The `toNumber` method calls `Math.abs` on the value passed to it ([see here](https://github.com/fnando/i18n-js/blob/master/app/assets/javascripts/i18n.js#L532)), this means that whatever value is passed to it is processed as a number. This causes issues when working with big numbers such as `123456789.123456789`, as JavaScript cannot handle large floating point numbers particularly well we lose precision.

For example `Math.abs("123456789.123456789")` returns `123456789.12345679`.

Is there a better solution we can use that doesn't require the value to be converted to a number first? It seems like this is only done in `i18n-js` to apply precision to a particular scale, if we don't supply a precision then could it just skip this line?

I'm happy to contribute, just looking for guidance or opinions.

PikachuEXE:
What if I make the default number options configurable?
Right now `NUMBER_FORMAT` is hardcoded and no way to change it.

PR #373, opened 2015-12-04
Fix rounding of unprecise floating points in I18n.toNumber()

kreintjes:
JavaScript's Float#toFixed method does not round all numbers correctly (or at least not how a human would expect). For example 0.105 is rounded as 0.10 instead of 0.11 when precision is 2. See this thread for more info: http://stackoverflow.com/questions/10015027/javascript-tofixed-not-rounding. Thanks to user2823670 for his improved toFixed implementation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 
 ### bug fixes
 - [Ruby] Fix of missing initializer at sprockets. ([#371](https://github.com/fnando/i18n-js/pull/371))
+- [JS] Correctly round unprecise floating point numbers.
 
 
 ## 3.0.0.rc11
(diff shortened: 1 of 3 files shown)


PikachuEXE:
I see a comment saying `Number 0.0000001 or less results in NaN`
I think a better fix for #365 is to avoid calling to `toFixed` with when precision is `null`.

kreintjes:
I wasn't trying to solve #365, I was trying to solve a different problem, namely the incorrect rounding of certain numbers. I was using I18n.toCurrency(0.105) which resulted in €0.10 instead of €0.11.

It seems that user2823670's shorter solution indeed does not handle very small numbers well. The full MDN solution (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/round) does not have this problem. I updated the pull request to implement this full solution instead.

PikachuEXE:
OK, this looks better now. 
Thanks for your contribution!
Merging

kreintjes:
You're welcome!

PR #374, opened 2015-12-08
Fix for Sprockets 3.0

mchristen:
If you are using Sprockets 3.0 w/ Rails the application will fail to boot with an error like

```
/home/matt/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/gems/i18n-js-3.0.0.rc11/lib/i18n/js/engine.rb:10:in `block in <class:Engine>': undefined method `register_preprocessor' for nil:NilClass (NoMethodError)
```

The following patch allows the application to boot up again.

### DIFF
--- a/lib/i18n/js/engine.rb
+++ b/lib/i18n/js/engine.rb
@@ -20,7 +20,9 @@ module I18n
         next unless JS::Dependencies.using_asset_pipeline?
         next unless JS::Dependencies.sprockets_supports_register_preprocessor?
 
-        Rails.application.assets.register_preprocessor "application/javascript", :"i18n-js_dependencies" do |context, source|
+        ##                    (sprockets 2.0 || sprockets 3.0)
+        sprockets = Rails.application.assets || Sprockets
+        sprockets.register_preprocessor "application/javascript", :"i18n-js_dependencies" do |context, source|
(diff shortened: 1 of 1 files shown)


PikachuEXE:
Is there any change log / PR / commit on rails/sprockets explaining the change on `Rails.application.assets`?
There is no test around this engine so I must be sure to know what should be used (supported by some evidence)

mchristen:
Check out the documentation for the options

```
config.assets.compile
```

```
config.assets.configure
```

on [rails/sprockets-rails](https://github.com/rails/sprockets-rails)

I'm not 100% sure if my fix is the correct fix, but using this gem w/o the above patch and with sprockets-rails master fails.

I found my fix through a gem that uses a similar setup, a custom preprocessor for sprockets. 

Checkout the history of the following file for some insight into when/why this code was introduced. 

https://github.com/reactjs/react-rails/commits/master/lib/react/rails/engine.rb

PikachuEXE:
Then should I be using `Sprockets` or `Rails.application.config.assets.configure` + block?

idlefingers:
@PikachuEXE I believe `Rails.application.config.assets.configure` is the correct approach

PikachuEXE:
Closing for #376

PR #375, opened 2015-12-17
Replace instanceof checks with their Underscore.js equivalents

TrevorBurnham:
This PR replaces all uses of `instanceof Object` and `instanceof Array` with the `_.isObject` and `_.isArray` functions from the latest Underscore.js release: https://github.com/jashkenas/underscore/blob/1.8.3/underscore.js#L1205-L1215 The reason is that `instanceof` can return `false` for objects and arrays passed between different JavaScript environments, e.g. [between a document and an iframe](http://tobyho.com/2011/01/28/checking-types-in-javascript/). In all other cases, these functions should be equivalent.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 - [Ruby] Fix of missing initializer at sprockets. ([#371](https://github.com/fnando/i18n-js/pull/371))
 - [Ruby] Use proper method to register preprocessor documented by sprockets-rails. ([#376](https://github.com/fnando/i18n-js/pull/376))
 - [JS] Correctly round unprecise floating point numbers.
+- [JS] Ensure objects are recognized when passed in from an iframe. ([#375](https://github.com/fnando/i18n-js/pull/375))
 
 
 ## 3.0.0.rc11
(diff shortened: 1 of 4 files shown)


PikachuEXE:
Thank for submitting this PR.
I have a few comments for this change:
- I can see underscore.js is trying to use `Array.isArray` first, then it's own implementation
- `toString` is an internal variable from underscore.js source code, but your version does not include it
- The fact that all tests passes means the test case for having array in `I18n.locales.get` is not tested.
  Please add test case for it.

TrevorBurnham:
@PikachuEXE Ah, that was sloppy of me! I've fixed the missing reference and added some test coverage.

PikachuEXE:
Looks fine
Can you add CHANGELOG entry & squash commits like:
- Add spec
- Change implementation
- Add CHANGELOG entry

TrevorBurnham:
Roger that, I've rebased and updated the CHANGELOG.

PikachuEXE:
Thanks!
Merging

PikachuEXE:
Pressed the wrong button :scream:

PR #376, opened 2015-12-18
Use Rails.application.config.assets.configure to load preprocessor

idlefingers:
This fixes issues with Rails.application.assets not being Sprockets in v3.

Related to https://github.com/fnando/i18n-js/pull/374


PikachuEXE:
Does it work with older sprockets-rails?

PikachuEXE:
It seems the document for 2.x is also suggesting to use the config block.
https://github.com/rails/sprockets-rails/blob/v2.3.3/README.md

Thanks for your PR!
Merging

Issue #285, opened 2014-11-26
Properly error if invalid key or missing placeholders given


MaryBartlett:
It would be really useful to throw an error if an invalid key or missing placeholder is given;
e.g. 
i18n.t('abc') where 'abc' doesn't exist - get error 'translation for 'abc' missing';
i18n.t('string with {{placeholder}}', {cat: 'dog'}); - get error 'placeholder missing, received {cat: 'dog'}'

At the moment, missing translations and placeholders do return a message that there has been an error, but unfortunately that message is a string so without further interpretation it isn't easy to see there's been an issue. 
"[missing "en-GB.abc" translation]"
"[missing {{placeholder}} value]"

johnnyshields:
Maybe this should be a console warning?

PikachuEXE:
How to log an error without breaking any browser compatibility?
I have zero knowledge for browser logging (of errors).
You know the famous IE...

johnnyshields:
One options would be to provide an onError callback into the I18n.js initializer. This could be then raised however the user feels appropriate.

TrevorBurnham:
I'd enjoy this feature as well. You could add an optional callback to `options`, e.g.

```js
I18n.missingTranslation = function(scope, options) {
  ...
  
  if (typeof options.onMissingTranslation === 'function') {
    options.onMissingTranslation(fullScopeWithLocale);
  }

  return '[missing "' + fullScopeWithLocale + '" translation]';
};
```

If you'd like, I can submit a pull request.

MaryBartlett:
Yes I'd be happy for you to take a look at this! I meant to when I raised it but had forgotten about it!

PR #380, opened 2015-12-21
Pass options object to nullPlaceholder and missingPlaceholder

TrevorBurnham:
`I18n.missingPlaceholder` exists so that it can be overridden for debugging purposes (#169). That's why it's passed `message`, even though the default implementation doesn't use it. This PR goes a small step further by also passing `options` to that function. This allows you to pass arbitrary debugging information through `I18n.translate` to `I18n.missingPlaceholder`, e.g. information about which part of the app failed to provide the placeholder.

Related: #285

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 
 ### enhancements
   - [JS] Allow extending of translation files ([#354](https://github.com/fnando/i18n-js/pull/354))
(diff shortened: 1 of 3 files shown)


PikachuEXE:
Thanks!
I think there should be some additional test cases for this change.
CHAGNELOG entry also :)

TrevorBurnham:
Sure thing. I've added a spec and CHANGELOG entry on this branch.

PikachuEXE:
Thanks!
Merging

PR #383, opened 2016-01-06
Throw an error when I18n.strftime() takes an invalid date

shinnn:
### Problem

JavaScript `Date` instance can be an `invalid date` in some cases. For example,

```js
// When `Date` constructor takes a valid date string
new Date('20001010'); //=> Tue Oct 10 2000 09:00:00 GMT+0900 (JST)

// When `Date` constructor takes an **invalid** date string
new Date('200010100'); //=> Invalid Date
```

Currently, I18n.strftime produces a wrongly formatted string when it takes an invalid date object. For example,

```js
I18n.strftime(new Date('foo'), '%Y%m%d');
//=> 'NaNaNaN'
```

### Solution

I updated `I18n.strftime` to immediately throw an error when it takes an `invalid date` object. This change prevents it from producing broken strings like `NaNaN`.

### DIFF
--- a/app/assets/javascripts/i18n.js
+++ b/app/assets/javascripts/i18n.js
@@ -772,6 +772,10 @@
 
     options = this.prepareOptions(options, DATE);
 
+    if (isNaN(date.getTime())) {
+      throw new Error('I18n.strftime() requires a valid date object, but received an invalid date.');
+    }
+
     var weekDay = date.getDay()
       , day = date.getDate()
       , year = date.getFullYear()
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Fixed
 
-- Nothing
+- [JS] Throw an error when `I18n.strftime()` takes an invalid date ([#383](https://github.com/fnando/i18n-js/pull/383))
+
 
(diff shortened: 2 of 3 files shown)


app/assets/javascripts/i18n.js
```diff
@@ -772,6 +772,10 @@
 
     options = this.prepareOptions(options, DATE);
 
+    if (isNaN(date.getTime())) {
+      throw new Error('I18n.strftime() requires a valid date object, but received an invalid date.');
```
PikachuEXE: Does `Error` work in every JS environment?

shinnn: Yes, `Error` constructor is one of the global object that all the JS environments (ancient browsers, Node.js, Electron etc.) have.

PikachuEXE:
Can you add an entry to CHANGELOG?
Then I will merge it.

shinnn:
> Can you add an entry to CHANGELOG?

https://github.com/fnando/i18n-js/commit/f0d4d31bac8c0c568bf7e120aff7bac733c38805#diff-4ac32a78649ca5bdd8e0ba38b7006a1eL17

PikachuEXE:
Thanks!
Merging

shinnn:
No problem! Thanks for your speedy review :)

PR #390, opened 2016-01-27
escape $ properly when interpolating

avetisk:


PikachuEXE:
Please provide test case that will fail with current implementaion first.

avetisk:
@PikachuEXE this is not a fix, this is simply a better way to write the same code :)

app/assets/javascripts/i18n.js
```diff
@@ -477,7 +477,7 @@
       name = placeholder.replace(this.placeholder, "$1");
 
       if (this.isSet(options[name])) {
-        value = options[name].toString().replace(/\$/gm, "_#$#_");
+        value = options[name].toString().replace(/\$/gm, "$$");
```
avetisk: Should have been '$$$$'

PR #401, opened 2016-04-07
Update link to NPM-Package to v3.0.0.rc12 from rc8

LinuCC:
The rc8-Version does not have the `.extend()` method implemented, so if you use the rails-gem as well as the NPM-package with rc8, Javscript complains that `.extend()` is missing.


PikachuEXE:
Thanks!

BTW I like your icon (and I know where is it from)

LinuCC:
Likewise :P Thank you for this great lib!

Issue #350, opened 2015-08-04
I18n::JS::Middleware autogeneration not working in Rails 4.2.3


seoyoochan:
I added `config.middleware.use I18n::JS::Middleware` in `development.rb`
but it does not work in rails 4.2.3
can someone reproduce this?

PikachuEXE:
Please provide more info first
such as versions of common gems like `rails` and `sprockets`

Describe what is expected and what actually happens

alvinsj:
@seoyoochan @PikachuEXE I have similar issue with the use of config/i18n-js.yml.
This is how I reproduce:

In `config/i18n-js.yml`:
```yaml
translations:
  - file: "public/javascripts/i18n/a.translations.js"
    only: ["*.a"]
  - file: "public/javascripts/i18n/b.translations.js"
    only: ["*.b"]
```
0. At first, everything is working. 
1. Remove what is in `public/javascripts/i18n/`
2. Refresh the app, and see `missing translations` error.
3. Remove `tmp/cache/i18n-js.yml`
4. Refresh the app again, It works now by newly generated files as specified in `config/i18n-js.yml`

PikachuEXE:
I guess the file is not checking whether the files exists or not.
I will change it later to accept additional options.

Thanks for providing the info. Middleware are hard to test (for me)

ezuk:
I just want to say that for me, this is not fixed. I just had to manually remove `tmp/cache/i18n-js.yml` to get it to work in dev mode.

PikachuEXE:
I plan to "fix" it before 3.0.0, but it might be be soon
If anyone are having this problem and able to fix it, please submit a PR so that other people with the same issue can test that :)

tleish:
It makes sense to me that the middleware cache would be cleared each time Rails starts.  I think checking to see if all the file still exists is the wrong approach because approach because which every page request you are parsing all the locale yaml files in order to get possible exported file names which would slow down each request.

If others feel they want the middleware to resolve with each page request, I think there should be a separate config setting in the 18n-js.yml.  Something like:

```yaml
ignore_middleware_cache: true
```

Again, I think this is the wrong approach and I've added a pull request for the approach that I think it the best strategy.

PikachuEXE:
@tleish 
Thanks for the PR
I just want to make sure no one is using that thing in production (or other places where performance is considered important)
Otherwise that caching can be removed (until a bug-less caching can be implemented)

tleish:
[...]

PR #402, opened 2016-04-13
Fix for #350 middleware caching

tleish:
clear middleware cache each time rails restarts and calls middleware is initialized which will resolve issue #350


PikachuEXE:
This is the same as removing the caching in middleware
Is that what you intended to do?

PikachuEXE:
Can you add a change log entry?

tleish:
Added entry to changelog

PikachuEXE:
Thanks :D

Issue #404, opened 2016-05-14
Translations do not update in rails 5 


steventen:
Hi, 
I'm using this gem in rails 5 rc1, and I'm using the Asset Pipeline setup.

In my local development, the translations couldn't update automatically after I update the content in locale yml files. I tried to restart the server and it won't work.

Right now, I have to manually run `rake tmp:clear` to make it work.

tilsammans:
Hi @steventen it's a known issue with this plugin #306 and `rake tmp:clear` is the current best way of dealing with this.

In production I actually update the asset version in order to expire all caches. It's the only way that works reliably for me.

AubinWinddle:
@tilsammans sorry but your response is false. This is only on Rails 5. It is due to the `using_asset_pipeline?` method which check only for Rails4 and Rails3. I will post a PR which cover this issue

tilsammans:
@AubinWinddle 👍

dbackeus:
We still have this problem after upgrading to Rails 5 even using rc14.

Have to clear tmp after every change to our translation yml files. Also production deploy does not build any new translation file.

PikachuEXE:
Besides rails 5 can you provide details about other asset related dependencies?
Like sprockets (with versions)

A more advanced (maybe not very easy for the first time but easier for debugging)
Is to checkout the `master`, change Gemfile to point to local path
Add debugging code (like `puts`) and see if there is anything output

But you need to clear the assets cache though, like `rake assets:clobber`
I the the above method when testing #418

dbackeus:
Tested with both `sprockets-rails` 3.1.1 and 3.2.0, `sprockets` is at 3.7.0.

Currently we've patched it by creating our own translations javascript file with explicit `depend_on` declaration for our languages ymls and then just adding in `I18n.translations = <%= I18n::JS.filtered_translations.to_json %>`.

If I get some time today will revert the hack and see if i18n-js master works out 👍

PikachuEXE:
Thanks
Let me know the result :)

dbackeus:
Ok got some info on this.

Apparently `JS::Dependencies.sprockets_supports_register_preprocessor?` is `nil` at the time of engine initialization due to `Sprockets` still being undefined (https://github.com/fnando/i18n-js/blob/master/lib/i18n/js/engine.rb#L12).

So it's a load order problem.

By explicitly adding `require "sprockets/railtie"` to `application.rb` before `Bundler.require(*Rails.groups)` it works out.

[...]

PR #413, opened 2016-07-15
Bugfix: make i18n-js compatible with rails 5

AubinWinddle:
I, this is a little fix to the open issue #404 to be able to use I18n-js with Rails 5 and asset pipeline. I think it will be more interesting to merge rails4 and rails5 methods since it's the same behavior. To discuss.

### DIFF
--- a/lib/i18n/js/dependencies.rb
+++ b/lib/i18n/js/dependencies.rb
@@ -12,6 +12,10 @@ module I18n
           safe_gem_check("rails", "~> 4.0", ">= 4.0.0.beta1") && running_rails4?
         end
 
+        def rails5?
+          safe_gem_check("rails", "~> 5.0", ">= 5.0.0.beta1") && running_rails5?
+        end
+
         def sprockets_supports_register_preprocessor?
           defined?(Sprockets) && Sprockets.respond_to?(:register_preprocessor)
         end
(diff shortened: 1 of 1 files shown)


PikachuEXE:
CI failed but should be related to Ruby vs rails version conflict since rails 5 requires Ruby 2.2.2+
Merging

guguso:
@pikachuExe could you please release a v3.0.0.rc14 version to make this fix available ?
Thks

PikachuEXE:
I will do it this week (today/tomorrow)
Sorry I planned to do it last Friday but was too busy

PikachuEXE:
Just released 3.0.0.rc14

guguso:
@PikachuEXE thks a lot

fwininger:
@PikachuEXE thanks !

SzNagyMisu:
@PikachuEXE would it be possible to update the _Installation_ part of the README file? I had the same issue, took me two hours to find that 3.0.0.rc11 is not the latest version. Thanks! And thanks for fixing the problem :-) !

PikachuEXE:
`gem "i18n-js", ">= 3.0.0.rc11"` should already make you install latest RC version
Unless it's already locked

SzNagyMisu:
I see. Then I'm not sure what really solved my problem. It seemed to me that it was the change from rc11 to rc15 which is the current latest according to [rubygems.org](https://rubygems.org/gems/i18n-js/versions). Thanks anyway.

fwininger:
I don't thinks that is the good place to open a thread to release version issue.

But to solve definitively this issue, could you use standard release number, like 3.0.15 and not 3.0.0.rc5. 

So everyone can use gem install i18-js or gem 'i18n-js'.

PikachuEXE:
Let's discuss further about version in another issue or gitter channel
https://gitter.im/fnando/i18n-js

PR #416, opened 2016-08-03
Support multiple translation files by merging objects

takashi:
When I have multiple locale files, `rake i18n:js:export` command will create multiple `.js` file from locale files.  

But in this situation with current implementation, i18n.js will overrides values of same key. 

for example: 

if there are files like below

```js
I18n.translations["en"] = I18n.extend((I18n.translations["en"] || {}), {"js": {"samples": { "hello": "world" }} })
```

```js
I18n.translations["en"] = I18n.extend((I18n.translations["en"] || {}), {"js": {"samples": { "i18n": "js" }} })
```

the result of `I18n.translations["en"]` that I expected is:

```
{
  "js": {
    "samples": {
      "hello": "world",
      "js": "i18n"
    }
  }
}
```

but I actually got:

```
{
  "js": {
    "samples": {
      "js": "i18n"
    }
  }
}
```

This is because `I18n.extend` method overrides the same key values. 

So I created `merge` method to fix this behavior. 

how about this?


PikachuEXE:
I guess you are trying to make it a deep merge?
`I18n.extend` is a function for shallow merge now.
I guess you can just change it to support deep merge.

But I am not sure if that will break things.

PikachuEXE:
Related to https://github.com/fnando/i18n-js/pull/354

takashi:
@PikachuEXE yes, I'd like to make it a deep merge. (and maybe I did in this PR?)
I guess the previous behavior seems to be supported with `merge` method of this PR as test shows.(but the test failed with installing activesupport..?)

If there is something I can fix, please tell me.

PikachuEXE:
You can ignore those "activesupport" build failures
Ruby/Rails version problem (build matrix not updated since rails 5 released)
Please add spec & doc for this behaviour change

And then we will ask for some people for opinion

takashi:
@PikachuEXE okey, thank for your advice. I'll add some specs and docs for this PR :)

takashi:
@PikachuEXE I added spec for deep merging and modify docs for this PR

PikachuEXE:
I invited people in #354 to comment on this PR
Let's wait a week to see if what opinion do they have.

takashi:
@PikachuEXE I waited a week and how about this PR? (seems no one leave comment on this PR ... ;) )
If this pr is ok, please merge this pr since i strongly wanted to use this feature with this gem

PikachuEXE:
I assume that no comment = no objection
Can you update https://github.com/fnando/i18n-js/blob/master/CHANGELOG.md?
I will merge after that
Thanks ;)

takashi:
@PikachuEXE  thanks :) 
I've updated changelog.md !

PikachuEXE:
Thanks!
Will merge tomorrow (leaving office now :P

PR #420, opened 2016-09-02
Allow numeric and boolean values in YAML

dan-brooks-otb:
When merging in values from YAML, I18n-js currently only treats string's and objects as valid values. If a key is set to a number or boolean it will not propagate through to i18n-js, and will instead return an object.

*YAML:*

```yml
currency:
  precision: 2
  delimiter: ','
  strip_insignificant_zeros: false
```

*Before:*

```js
{
  currency: {
    precision: {},
    delimiter: ',',
    strip_insignificant_zeros: {}
  }
}
```

*After:*

```js
{
  currency: {
    precision: 2,
    delimiter: ',',
    strip_insignificant_zeros: false
  }
}
```

### DIFF
--- a/app/assets/javascripts/i18n.js
+++ b/app/assets/javascripts/i18n.js
@@ -53,11 +53,19 @@
 
   // Is a given value an array?
   // Borrowed from Underscore.js
-  var isArray = function(obj) {
+  var isArray = function(val) {
     if (Array.isArray) {
-      return Array.isArray(obj);
+      return Array.isArray(val);
     };
-    return Object.prototype.toString.call(obj) === '[object Array]';
+    return Object.prototype.toString.call(val) === '[object Array]';
+  };
+
+  var isString = function(val) {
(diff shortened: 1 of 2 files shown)


PikachuEXE:
Thanks
Sorry for late response
I am quite busy this month

n0nick:
Thanks @dan-brooks-otb @PikachuEXE for the quick catch and handling! We were scratching our heads about this one.

@PikachuEXE any chance this could be released to rubygems.org (rc15?)? I'd rather avoid Github dependencies on Gemfiles...

PikachuEXE:
Sure, will do it soon (this week or next)

n0nick:
ping 😊

PikachuEXE:
Will do so when #436 fixed, sorry for delaying :S

PikachuEXE:
In case you don't know a new version is released
@n0nick

PR #421, opened 2016-09-03
Make it clear that fallbacks must be set to false for the single-locale config option to work.

hoffmanc:


PikachuEXE:
If `fallback` not set to false what will happen?

johnnyshields:
This issue should be closed. Fallbacks doesn't need to be false; if it's true then you'll get the standard fallbacks behavior which is fine.

johnnyshields:
@PikachuEXE

PR #422, opened 2016-09-06
Fix for merge so that numbers don't get turned into {}'s

mipearson:
Since upgrading to the latest version of i18n-js we started having issues with `I18n.toNumber` - it would always result in `NaN`.

Debugging it showed that our `numbers.format` localisation was actually coming through I18n as:
image

Where what we expected was:
image

This was traced back to the merge method in i18n.js. It treated strings correctly, but not  booleans/numbers/array. A simple test case for this follows:

```javascript
I18n.extend({cats: 1}, {cats: 1})
// {cats: 1}
I18n.extend({}, {cats: 1})
// {cats: {}}
```

### DIFF
--- a/app/assets/javascripts/i18n.js
+++ b/app/assets/javascripts/i18n.js
@@ -83,7 +83,7 @@
     var key, value;
     for (key in obj) if (obj.hasOwnProperty(key)) {
       value = obj[key];
-      valueType = Object.prototype.toString.call(value);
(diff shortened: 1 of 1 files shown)


PikachuEXE:
I just merged #420 which seems similar to this PR
Maybe you can try `master` and see if that's fixed?
Thanks.

dnl:
#420 fixes it, thanks :)

Issue #423, opened 2016-09-16
Should fallback locales be validated?


coremessage:
Rails does not validate fallback locales, but i18n-js does. Do we have to perform this check? Should we make this configurable? Absence of locale validations can be useful for setting up locale aliases without having to pollute `I18n.available_locales`.

PikachuEXE:
Where does the validation occur?
Can you also give an example of the usage you explained?
I am a helper so not very familierized with every bit of this gem.
Thanks.

coremessage:
The validation occurs [here](https://github.com/fnando/i18n-js/blob/master/lib/i18n/js/fallback_locales.rb#L35).

As an example, I want my Rails app to recognize "no" country code as Norway, but all Norwegian translations in rails are stored in "nn" locale. (Same goes for some other languages)
To set up an alias, I configure fallback locales like this:
```ruby
# config/application.rb
# Most of the code omitted
config.i18n.available_locales = ['en', 'no']
config.i18n.fallbacks = { no: :nn }
```
This essentially tells Rails to search for "no" translations in "nn.yml". Then, I configure i18n-js:
```coffeescript
# app/assets/javascripts/translations.coffee
# Most of the code omitted
I18n.fallbacks = true
```
However, i18n-js requires "nn" to be included into `I18n.available_locales`, which makes no sense since "no" and "nn" are the same locale in this case. This can be a problem if your app uses `I18n.available_locales` for a language switcher (which is natural).

Why do we have to perform the check, anyway?

PikachuEXE:
I guess it's for safety at first
But if there is a valid case I am happy to remove it (since rails/I18n itself don't care as well)
If you want the change to be made faster you can submit a PR ;)

coremessage:
Thank you guys! Should I close the issue?

PikachuEXE:
If you have tested it (master) and no problem found
Then please close :)

coremessage:
Okay, locale validation is gone, but now the issue is i18n-js still reads translations from files specified in `I18n.available_locales`. This means that translations for languages that are present in `config/locales`, but not specified in `I18n.available_locales`, are not loaded. It seems that Rails, however, loads locales from `I18n.backend.available_locales`. I've put a pull request together to add an option to override where i18n-js fetches translations from.

Btw, should we make `I18n.backend.available_locales` the default instead of `I18n.available_locales`?

PikachuEXE:
Do you know where does rails call `I18n.backend.available_locales`?
[...]

PR #425, opened 2016-09-17
Remove fallback locale validation

akxcv:
(From issue #423)

Whatever the use case, if we want i18n-js to behave more like Rails, we shouldn't check if all fallback locales are specified under `I18n.available_locales`, because Rails doesn't.


PikachuEXE:
LGTM :+1:

PR #426, opened 2016-09-20
Add an option to specify where to fetch translations from

coremessage:
See #423

### DIFF
--- a/lib/i18n/js/segment.rb
+++ b/lib/i18n/js/segment.rb
@@ -3,24 +3,25 @@ module I18n
 
     # Class which enscapulates a translations hash and outputs a single JSON translation file
     class Segment
-      OPTIONS = [:namespace, :pretty_print, :js_extend, :sort_translation_keys].freeze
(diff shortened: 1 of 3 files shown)


PikachuEXE:
Is this still needed?
If so I will create another PR for the issue
But using `eval` is too much...

lib/i18n/js/segment.rb
```diff
@@ -3,24 +3,25 @@ module JS
...
+        @js_extend     = options.key?(:js_extend) ? !!options[:js_extend] : true
         @sort_translation_keys = options.key?(:sort_translation_keys) ? !!options[:sort_translation_keys] : true
+        @locale_source = options[:locale_source] || "I18n.available_locales"
       end
 
       # Saves JSON file containing translations
       def save!
         if @file =~ LOCALE_INTERPOLATOR
-          I18n.available_locales.each do |locale|
+          eval(@locale_source).each do |locale|
```
johnnyshields: eval is evil, it can be used to execute arbitrary code. instead, define allowed formats for this option and implement a parsing function

PR #427, opened 2016-09-23
Fix/missing dependency

PikachuEXE:
This PR removes development dependency `activesupport`
Which is actually a hidden runtime dependency
since `with_indifferent_access` and `slice` are used

`with_indifferent_access` is from `activesupport`
`slice` is from `activesupport` 
(although `i18n` also "provide" `slice`, but it won't be used if `activesupport` is loaded first)

The test pass but not sure if it's working properly for everyone

### DIFF
--- a/i18n-js.gemspec
+++ b/i18n-js.gemspec
@@ -20,7 +20,6 @@ Gem::Specification.new do |s|
 
   s.add_dependency "i18n", "~> 0.6", ">= 0.6.6"
   s.add_development_dependency "appraisal", "~> 2.0"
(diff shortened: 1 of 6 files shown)


PikachuEXE:
This also fix some Travis bulld due to incompatibility with latest `activesupport` and some older Ruby versions

PikachuEXE:
@johnnyshields 
Do you have any comment on this change?

johnnyshields:
@PikachuEXE I'd rename namespace `::Private` to `::Support`, otherwise looks fine.

PR #429, opened 2016-09-24
Add available_locales config option

akxcv:
This option allows to specify locales to use in Javascript. Defaults to `I18n.available_locales`. See #423

### DIFF
--- a/lib/i18n/js.rb
+++ b/lib/i18n/js.rb
@@ -60,7 +60,7 @@ module I18n
 
     # deep_merge! given result with result for fallback locale
     def self.merge_with_fallbacks!(result)
-      I18n.available_locales.each do |locale|
+      available_locales.each do |locale|
         fallback_locales = FallbackLocales.new(fallbacks, locale)
         fallback_locales.each do |fallback_locale|
           # `result[fallback_locale]` could be missing
@@ -149,10 +149,13 @@ module I18n
 
     # Initialize and return translations
     def self.translations
-      ::I18n.backend.instance_eval do
+      backend_translations = ::I18n.backend.instance_eval do
         init_translations unless initialized?
-        translations.slice(*::I18n.available_locales)
+        translations
       end
+
+      return backend_translations if available_locales.empty?
+      backend_translations.slice(*available_locales)
     end
 
     def self.use_fallbacks?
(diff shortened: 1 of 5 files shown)


lib/i18n/js.rb
```diff
@@ -188,6 +191,13 @@ def self.extract_segment_options(options)
       segment_options.merge(options.slice(*Segment::OPTIONS))
     end
 
+    def self.available_locales
+      config.fetch(:available_locales) do
+        # default value
+        I18n.available_locales rescue []
```
PikachuEXE: Why `rescue` is used here?

lib/i18n/js.rb
```diff
@@ -149,10 +149,13 @@ def self.filter(translations, scopes)
...
     # Initialize and return translations
     def self.translations
-      ::I18n.backend.instance_eval do
+      backend_translations = ::I18n.backend.instance_eval do
         init_translations unless initialized?
-        translations.slice(*::I18n.available_locales)
+        translations
       end
+
+      return backend_translations if available_locales.empty?
```
PikachuEXE: Why do we want to return all backend translations when `available_locales` is empty?
We should remind them or at least return nothing (so developers notice they might have set something wrong)

PikachuEXE:
I just merged #427 so some files contain conflicts now
Let me know if you are not sure how to fix
Thanks

lib/i18n/js/segment.rb
```diff
@@ -3,24 +3,25 @@ module JS
...
-        @namespace    = options[:namespace] || 'I18n'
-        @pretty_print = !!options[:pretty_print]
-        @js_extend    = options.key?(:js_extend) ? !!options[:js_extend] : true
+        @file                  = file
+        @translations          = translations
+        @namespace             = options[:namespace] || 'I18n'
+        @pretty_print          = !!options[:pretty_print]
+        @js_extend             = options.key?(:js_extend) ? !!options[:js_extend] : true
         @sort_translation_keys = options.key?(:sort_translation_keys) ? !!options[:sort_translation_keys] : true
+        @available_locales     = options[:available_locales] || I18n.available_locales rescue []
```
akxcv: Yeah, it's definitely duplication, but it's required in `segment`, too, see line 23.

lib/i18n/js.rb
akxcv: If we get a `MethodMissing` (`I18n.available_locales`), we default to an empty array.

akxcv:
Yeah, I'll fix the conflicts once we decide on the best way to add the `available_locales` option.

akxcv:
Closing because no activity for several months

PR #433, opened 2016-11-10
Add fallback to pluralize

xykox:
When fallbacks activated, Pluralization won't work :

```javascript
I18n.fallbacks = "en";
I18n.locale = "en";
I18n.t("inbox.counting"); // { zero: "No message", one: "You a message", other: "You have %{count} messages"}
I18n.t("inbox.counting", {count: 10}); // You have 10 messages

I18n.locale = "fr";
I18n.t("inbox.counting", {count: 10}); // [missing "fr.inbox.counting.other" ]
// Expect english fallback : You have 10 messages
```

### DIFF
--- a/app/assets/javascripts/i18n.js
+++ b/app/assets/javascripts/i18n.js
@@ -347,6 +347,8 @@
       , locale
       , scopes
       , translations
+      , keys
+      , key
     ;
 
     scope = this.getFullScope(scope, options);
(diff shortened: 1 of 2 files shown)


PikachuEXE:
My question would be:
if the translation is:
```
I18n.translations["pt-BR"].inbox= {
       one: "something"
     , other: null
     , zero: "something"
   };
```
What would `I18n.t("inbox.counting", {count: 10});` return?

xykox:
Thanks for your feedback. Actually it will return :
[missing "pt-BR.inbox.other" translation]

So I will try modify the code to have the fallback instead of the "missing" message.

PikachuEXE:
Can you make whatever changes you need (if any left) and then squash commits?
Will merge after that

xykox:
Added some specs to meet requirements.
I completely rewrote my part.
Hope it suits

PikachuEXE:
Thanks!

PR #449, opened 2017-04-05
Feature/lazy default message

Rafe:
Hi,

This pull request add a lazy evaluation feature to default message in translation,

ex:

```
I18n.t('foo.bar', {
  defaultValue: function(scope) {
    return scope.toUpperCase()
  }
})
// FOO.BAR
```

Also do some refactor to clear out where to put the lazy evaluation logic.

### DIFF
--- a/app/assets/javascripts/i18n.js
+++ b/app/assets/javascripts/i18n.js
@@ -52,7 +52,12 @@
   // Borrowed from Underscore.js
   var isObject = function(obj) {
     var type = typeof obj;
-    return type === 'function' || type === 'object' && !!obj;
+    return type === 'function' || type === 'object'
+  };
+
+  var isFunction = function(func) {
+    var type = typeof func;
+    return type === 'function'
   };
 
   // Check if value is different than undefined and null;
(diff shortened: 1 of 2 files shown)


PikachuEXE:
Thanks for your contribution. The spec looks fine.

However the change set is too large for just adding this feature.
I understand you want to refactor **AND** add this feature too.
But having both in the same PR makes it harder and more time-consuming to be reviewed.

Please submit this as a refactor or as feature addition.

Rafe:
Fixed, separate the refactor part to another pull request, and thanks for reviewing.

app/assets/javascripts/i18n.js
```diff
@@ -55,6 +55,11 @@
     return type === 'function' || type === 'object' && !!obj;
   };
 
+  var isFunction = function(func) {
+    var type = typeof func;
+    return type === 'function' && !!func;
```
PikachuEXE: What's the purpose of `!!func`?

Rafe: check if func exists, but it is probably not necessary here since `typeof` already block other values. Will remove the `!!func` check

PikachuEXE:
@Rafe Got a merged conflict
Please solve it so I can review this PR :)

Rafe:
conflict fixed

PikachuEXE:
Will look at another PR tomorrow :)

PR #450, opened 2017-04-06
Refactor/remove side effect and prepare options

Rafe:
Remove the code that update `options`,

It makes the code harder to read because we need to consider the case that options.defaultMessage and options.count will change after function call.

And generally, keep object immutable is a better practice.

### DIFF
--- a/app/assets/javascripts/i18n.js
+++ b/app/assets/javascripts/i18n.js
@@ -321,7 +321,7 @@
   // This is used internally by some functions and should not be used as an
   // public API.
   I18n.lookup = function(scope, options) {
-    options = this.prepareOptions(options);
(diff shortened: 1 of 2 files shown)


PikachuEXE:
@Rafe 
Please solve the merge conflict thx :)

Rafe:
conflict fixed

PikachuEXE:
@Rafe Got one more conflict to solve lol
I guess it's due to the extracted PRs merged

Rafe:
Fixed, just small conflict in spec file, I should rebase after branch got merged.

PikachuEXE:
@Rafe 
Merged, thx :)
I wonder if this should be mentioned in change log...

PR #451, opened 2017-04-06
Refactor/simplify initializer

Rafe:
Simplify initializer logic

### DIFF
--- a/app/assets/javascripts/i18n.js
+++ b/app/assets/javascripts/i18n.js
@@ -55,6 +55,11 @@
     return type === 'function' || type === 'object' && !!obj;
   };
 
+  // Check if value is different than undefined and null;
+  var isSet = function(value) {
+    return typeof(value) !== 'undefined' && value !== null;
+  };
+
   // Is a given value an array?
   // Borrowed from Underscore.js
   var isArray = function(val) {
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/i18n.js
```diff
@@ -55,6 +55,11 @@
     return type === 'function' || type === 'object' && !!obj;
   };
 
+  // Check if value is different than undefined and null;
+  var isSet = function(value) {
+    return value !== undefined && value !== null;
+  }
```
PikachuEXE: missing semicolon

app/assets/javascripts/i18n.js
```diff
@@ -55,6 +55,11 @@
     return type === 'function' || type === 'object' && !!obj;
   };
 
+  // Check if value is different than undefined and null;
+  var isSet = function(value) {
+    return value !== undefined && value !== null;
```
PikachuEXE: I think the common check should be `typeof value !== "undefined" && value !== null`
This avoids accident when `undefined` assigned to something else

Rafe: fixed

PikachuEXE:
Thanks!
Will also look at other PR soon

Issue #303, opened 2015-02-06
Rails 3, doesn't seem to pickup "new" stuff I put in en.yml


rstudner:
Okay.. very strange.

Using the library, everything was working great.

All of a sudden, if I add things to en.yml they don't show up in the I18n.translations.

It seems to be fixed in time/stuck and i'm not sure how to kick it :)

rstudner:
I ran a bundle exec rake assets:precompile and now I appear to be seeing my updates....

But if I make 'newer' updates .. it again doesn't seem them.  Hrmph.

PikachuEXE:
Are you talking about in development or production?
Is `rake assets:precompile` run on deployment after `en.yml` is modified? (if production)

gmccullough:
I believe I'm running into the same issue. Any edits to config/locals/*.yml aren't reflected in filtered.js unless I stop my server, run rake assets:clean, and restart the server. Seems like this shouldn't be necessary. This is in development.

I just started to use this gem and ran into this issue, so I may have some sort of misconfiguration on my end. Any suggestions?

PikachuEXE:
Which version of the gem are you using?
All README is for the `rc` versions of the gem, not official version will be released soon :S

rstudner:
I'm using the 0.12.2/latest stable gem at the moment.  building the 1.0.0 pre8 one right now to try it as well.

gmccullough:
I'm using 3.0.0.rc8

jesperronn:
@rstudner @gmccullough in #306 I learned to run `rake tmp:cache:clear`, which finally helped me reload the translations. I also tried `rake assets:clobber` and `rm public/javascripts/translations.js` and updating my yaml files but that was in vain.

I have now set our ci-build up to clear cache before it precompiles assets with `rake tmp:cache:clear`.

PR #453, opened 2017-04-13
Fix for rails 3 not picking up new stuff 

artygus:
Addresses issue #303 Rails 3, doesn't seem to pickup "new" stuff I put in en.yml


lib/i18n/js/engine.rb
```diff
@@ -9,56 +10,58 @@ class SprocketsExtension
     end
 
     class Engine < ::Rails::Engine
-      if JS::Dependencies.sprockets_supports_register_preprocessor?
```
PikachuEXE: If you remove this comment it might cause some errors (when sprockets lib absent)
This is why we are using `sprockets_supports_register_preprocessor?`

lib/i18n/js/engine.rb
```diff
@@ -9,56 +10,58 @@ class SprocketsExtension
...
+        # https://github.com/rails/sprockets-rails/blob/v3.0.0/README.md
+        if JS::Dependencies.sprockets_supports_register_preprocessor?
           Rails.application.config.assets.configure do |config|
             config.register_preprocessor(
               "application/javascript",
               ::I18n::JS::SprocketsExtension,
             )
           end
+        else
+          Rails.application.assets.register_preprocessor "application/javascript", ::I18n::JS::SprocketsExtension
```
PikachuEXE: if `sprockets_supports_register_preprocessor?` is `false`
Should we even try to call `Rails.application.assets.register_preprocessor`?

lib/i18n/js/engine.rb
artygus: Oh yeah, now I see. Btw even if sprockets is here, blank rails 3 app doesn't have "sprockets/version" required. I'll craft a workaround

PikachuEXE: I guess you need to require `sprockets/version` with `rescue LoadError`?
Again sprockets might be absent ~_~

lib/i18n/js/engine.rb
artygus: Yeah, latest version of sprockets supported by rails 3 is 2.2 and it doesn't expose static register_preprocessor method to Sprockets class. 

`Rails.application.config.assets` is an instance of `Sprockets::Railtie::OrderedOptions` in Rails 4, while it's simply `ActiveSupport::OrderedOptions` in rails 3, so calling configure does nothing for rails 3

lib/i18n/js/dependencies.rb
```diff
@@ -16,8 +16,8 @@ def rails5?
           safe_gem_check("rails", "~> 5.0", ">= 5.0.0.beta1") && running_rails5?
         end
 
-        def sprockets_supports_register_preprocessor?
-          defined?(Sprockets) && Sprockets.respond_to?(:register_preprocessor)
+        def sprockets_rails?
+          safe_gem_check("sprockets-rails") && !rails3?
```
artygus: there's a 0.0.1 version of the gem that could be installed with rails 3

lib/i18n/js/engine.rb
artygus: does it look better now?

PikachuEXE:
Thanks
You can test it by using master of this repo

PR #458, opened 2017-06-13
Update link to CLDR plural rules

bahrieinn:
Current link no longer contains plural rules and has a generic link to CDLR homepage.


PikachuEXE:
Thanks for updating that!

PR #465, opened 2017-07-20
Fix 3 part locale does not fallback to 2 part locale

### LINKED ISSUE #460: Fallback bug
I forward this issue from https://github.com/AlexanderZaytsev/react-native-i18n/issues/96#event-1128148909.

----

The fallback strategy has a bug. The Chinese mandarin language (zh/中, how to pronounce) use a character system (Han/漢, how to write) which is divided into two kinds of symbol styles: [Traditional Chinese characters (**zh-Hant**, where **t** stands for **T**raditional)](https://en.wikipedia.org/wiki/Traditional_Chinese_characters) and [Simplified Chinese characters (**zh-Hans**, where **s** stands for **S**implified)](https://en.wikipedia.org/wiki/Simplified_Chinese_characters).

For example, a user uses Traditional Chinese characters and lives in Taiwan, his language and locale should be `zh-Hant-TW`, and the priority of the i18n fallback strategy should be:

1. `zh-Hant-TW`: Traditional Chinese characters used in Taiwan
2. `zh-Hant`: general Traditional Chinese characters
3. `zh`: general Chinese characters

The current fallback strategy of `i18n-js` has a bug. If it doesn't find `zh-Hant-TW`, it will only look for `zh`, not `zh-Hant` first.

I want to provide `zh-Hant` for all Traditional Chinese and `zh-Hans` for all Simplified Chinese no matter what the locale is, but it can't currently work. Devices provide `lang+locale` and `i18n-js` fallbacks to `zh` directly in this scenario.

Fallback strategy:

* Expected: `zh-Hant-TW` -> `zh-Hant` -> `zh` -> default language
* Current: `zh-Hant-TW` -> `zh` -> default language

```js
// general Chinese
zh

// Traditional Chinese 
zh-Hant

// Simplified Chinese
zh-Hans

// Traditional Chinese + Locale
zh-Hant-TW
zh-Hant-SG
zh-Hant-HK
zh-Hant-CN

// Simplified Chinese + Locale
zh-Hans-TW
zh-Hans-SG
zh-Hans-HK
zh-Hans-CN
```

PikachuEXE:
Fix #460

### DIFF
--- a/app/assets/javascripts/i18n.js
+++ b/app/assets/javascripts/i18n.js
@@ -246,8 +246,6 @@
   I18n.locales["default"] = function(locale) {
     var locales = []
       , list = []
-      , countryCode
-      , count
     ;
 
     // Handle the inline locale option that can be provided to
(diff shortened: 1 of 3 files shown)


PikachuEXE:
Merged since no one objects this change

PR #467, opened 2017-07-25
Simplify Rails version checks

### LINKED ISSUE #448: i18n:js:export task not exposed in Rails
It seems like the `i18n:js:export` task is not available anymore when including this Gem in a Rails (4) project. Possibly since the big refactor that removed the Railtie.

The [task is still implemented](https://github.com/fnando/i18n-js/blob/b8fb918c827a2d3bbe187372e47bbe1a7a6c4c19/lib/tasks/export.rake), but it is not registered anywhere. Or am I missing some magic Rails convention?

### LINKED ISSUE #466: Why check for Rails gem to be installed?
Believe it or not, our Rails project does not actually require Rails in the Gemfile. Instead, we require the various sub-frameworks manually, because we don't need them all (notably excluding activerecord). Don't ask, it's a long story...

This means that i18n-js 3.0 will not load properly in our project.

It took me a while to track this down, but I finally found that you are checking not only for `Rails` to be available and the appropriate version to be returned from `Rails.version`. You are also checking for the `rails` gem to be listed in the Gemfile. For us, that obviously causes problems.

IMO (and because it would fix our problem), checking for the correct Rails version to be available in the runtime should surely be enough? You don't need to bother where it comes from, just that it's there.

By the way: seeing that that code [was apparently inspired by similar code in the jasmine-gem](https://github.com/fnando/i18n-js/commit/07978b0558c30866c8e6640dcd14aed2db332e5b), you might be interested to know that [the jasmine-gem removed that code again](https://github.com/jasmine/jasmine-gem/commit/c726fe1e93ae9848190d7f10901dc782ee0aebc0) to fix other problems with the Gem version resolution.

---

I will prepare a pull request, but wanted to understand this first. Maybe there is some hidden necessity for this code to be in place? :smile: 

P.S.: Relevant XKCD reference: https://xkcd.com/1172 :wink: 
P.P.S.: This may very well be the underlying issue for #448.

franzliedke:
There is no need to check for the rails gem to be installed.
We are only concerned with whether Rails is available and
don't care where it comes from (as there are other ways to
install Rails).

Fixes #466.

I just tried it out and this also resolves #448.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Changed
 
-- Nothing
+- [Ruby] Relax Rails detection code to work with alternative installation methods
+  (PR: https://github.com/fnando/i18n-js/pull/467)
 
 ### Fixed
 
--- a/lib/i18n/js/dependencies.rb
+++ b/lib/i18n/js/dependencies.rb
@@ -4,28 +4,12 @@ module I18n
     # we need to specify pre-release version suffix in version constraint
     module Dependencies
       class << self
-        def rails3?
-          safe_gem_check("rails", "~> 3.0") && running_rails3?
-        end
-
-        def rails4?
-          safe_gem_check("rails", "~> 4.0", ">= 4.0.0.beta1") && running_rails4?
-        end
-
-        def rails5?
-          safe_gem_check("rails", "~> 5.0", ">= 5.0.0.beta1") && running_rails5?
+        def rails?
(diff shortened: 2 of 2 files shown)


PikachuEXE:
Good idea!
What would you suggest to write in change log entry?
If possible just add it before it's merged. Thanks.

franzliedke:
I updated the changelog. :)

PikachuEXE:
Great!
I am gonna wait a few days for #465 and merge it if no one objects that PR
And release after that

franzliedke:
Thanks for the release! It works for us now. :heart:

PR #472, opened 2017-09-18
Fix merging of plural keys

johnnyshields:


lib/i18n/js/utils.rb
```diff
@@ -1,10 +1,16 @@
 module I18n
   module JS
     module Utils
+      PLURAL_KEYS = %i[zero one two few many other]
```
PikachuEXE: Where does this come from?
Also it's not `frozen?`

lib/i18n/js/utils.rb
```diff
@@ -1,10 +1,16 @@
...
+
       # deep_merge by Stefan Rusterholz, see <http://www.ruby-forum.com/topic/142809>.
       # The last result is modified to treat `nil` as missing key
       MERGER = proc do |_key, v1, v2|
-        Hash === v1 && Hash === v2 ? v1.merge(v2, &MERGER) : (v2.nil? ? v1 : v2)
+        if Hash === v1 && Hash === v2
+          (v2.keys - PLURAL_KEYS).empty? ? v2 : v1.merge(v2, &MERGER)
```
PikachuEXE: Please add some code comment to explain the code here
Also please add test case

johnnyshields:
This fixes the case that the key being merged is a pluralization. Suppose I have:

```
en:
  apple:
    one: Apple
    other: Apples

fr:
  apple:
    zero: Pomme
    one: Pomme
    other: Pommes
```

With fallbacks, this will merge to:

```
en:
  apple:
    zero: Pomme
    one: Apple
    other: Apples
```

PikachuEXE:
Test case please 0w0

PikachuEXE:
Let me know when test cases are added

johnnyshields:
@PikachuEXE this is ready to merge

johnnyshields:
@PikachuEXE this is ready to merge

ghiculescu:
@johnnyshields this broke things in our app. Can you please confirm if https://github.com/fnando/i18n-js/pull/551 works for you.

johnnyshields:
#551 does not work for me; see comment there. Recommend you to fix your app.

PR #473, opened 2017-09-21
Avoid writing files if we don't have to.

mipearson:
Writing files triggers other actions in anything watching those files - eg
rebuilding a webpack bundle. Normally this won't occur as i18n-js's caching
will determine that nothing needs to be done, but it can add up quickly
when running individual specs from the command line, where the cache
is reset on every load.

### DIFF
--- a/lib/i18n/js.rb
+++ b/lib/i18n/js.rb
@@ -219,6 +219,9 @@ module I18n
         FileUtils.mkdir_p(export_i18n_js_dir_path)
 
         i18n_js_path = File.expand_path('../../../app/assets/javascripts/i18n.js', __FILE__)
+        destination_path = File.expand_path("i18n.js", export_i18n_js_dir_path)
+        return if File.exist?(destination_path) && File.identical?(i18n_js_path, destination_path)
+
         FileUtils.cp(i18n_js_path, export_i18n_js_dir_path)
       end
 
--- a/lib/i18n/js/segment.rb
+++ b/lib/i18n/js/segment.rb
@@ -40,11 +40,15 @@ module I18n
 
       def write_file(_file = @file, _translations = @translations)
         FileUtils.mkdir_p File.dirname(_file)
+        contents = js_header
+        _translations.each do |locale, translations_for_locale|
+          contents << js_translations(locale, translations_for_locale)
+        end
+
+        return if File.exist?(_file) && File.read(_file) == contents
+
         File.open(_file, "w+") do |f|
-          f << js_header
-          _translations.each do |locale, translations_for_locale|
-            f << js_translations(locale, translations_for_locale)
-          end
+          f << contents
         end
       end
 


lib/i18n/js.rb
```diff
@@ -219,6 +219,9 @@ def self.export_i18n_js
         FileUtils.mkdir_p(export_i18n_js_dir_path)
 
         i18n_js_path = File.expand_path('../../../app/assets/javascripts/i18n.js', __FILE__)
+        destination = File.expand_path("i18n.js", export_i18n_js_dir_path)
+        return if File.exist?(destination) && FileUtils.cmp(i18n_js_path, destination)
```
PikachuEXE: Maybe use `identical?` instead of `cmp` since it's a clearer method name?

lib/i18n/js.rb
```diff
@@ -219,6 +219,9 @@ def self.export_i18n_js
         FileUtils.mkdir_p(export_i18n_js_dir_path)
 
         i18n_js_path = File.expand_path('../../../app/assets/javascripts/i18n.js', __FILE__)
+        destination = File.expand_path("i18n.js", export_i18n_js_dir_path)
```
PikachuEXE: Better name: `destination_path`

lib/i18n/js.rb
```diff
@@ -219,6 +219,9 @@ def self.export_i18n_js
...
+        return if File.exist?(destination) && FileUtils.cmp(i18n_js_path, destination)
+
         FileUtils.cp(i18n_js_path, export_i18n_js_dir_path)
```
PikachuEXE: Maybe we should also change `export_i18n_js_dir_path ` to `destination`?

mipearson:
@PikachuEXE Will do - clocking off for today. Thanks for the speedy review! :)

mipearson:
I've changed a couple of the variable names as requested.

I haven't changed `export_i18n_js_dir_path` as I'm not sure what repercussions that will have for the rest of the codebase.

I tried adding tests, but doing so effectively would probably require a rewrite of the specs they're very strongly based around saving the files in the `before` blocks.

PikachuEXE:
Then at least test it manually once :P

mipearson:
I've tested this on our local codebase. Or do you mean writing a new test, separate from the existing unit tests?

PikachuEXE:
I mean test this change in your code that uses it.
If it fails later we can write test cases for it later (let's hope not)

lib/i18n/js.rb
```diff
@@ -219,6 +219,9 @@ def self.export_i18n_js
         FileUtils.mkdir_p(export_i18n_js_dir_path)
 
         i18n_js_path = File.expand_path('../../../app/assets/javascripts/i18n.js', __FILE__)
+        destination_path = File.expand_path("i18n.js", export_i18n_js_dir_path)
+        return if File.exist?(destination_path) && File.identical?(i18n_js_path, destination_path)
```
franzliedke: Is this one supposed to compare file contents as well? As far as I can tell [from the docs](https://ruby-doc.org/core-2.2.3/File.html#method-c-identical-3F), `identical?` does not check the file contents, but only for links etc.

PikachuEXE: [Which is fixed afterward](https://github.com/fnando/i18n-js/commit/a88b206644a14b0de7bbb58e0ba1baab93737a8e)

PR #474, opened 2017-09-25
Replace File.identical with FileUtils.compare_file

mipearson:
File.identical? only tells you if it's the same actual file (eg, a symlink or hardlink), but returns false if two different files have the same contents.

I've confirmed this change works - I'd confirmed the original version with `.cmp` (alias for `compare_file`), but not with `.identical?`.

### DIFF
--- a/lib/i18n/js.rb
+++ b/lib/i18n/js.rb
@@ -220,7 +220,7 @@ module I18n
 
         i18n_js_path = File.expand_path('../../../app/assets/javascripts/i18n.js', __FILE__)
         destination_path = File.expand_path("i18n.js", export_i18n_js_dir_path)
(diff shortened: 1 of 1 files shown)


PikachuEXE:
I missed that while reviewing!
Fixed in a88b206644a14b0de7bbb58e0ba1baab93737a8e

I actually mean [`FileUtils.identical?`](https://ruby-doc.org/stdlib-2.4.0/libdoc/fileutils/rdoc/FileUtils.html#method-c-identical-3F) which is just an alias to `FileUtils.compare_file`

PikachuEXE:
@mipearson 
Let me know when `master` tested
I will release after that

mipearson:
I .. forgot about this. You've already released, so I'll close this PR. Thankyou :)

PR #488, opened 2018-01-24
Fix a bug when rails project has only Webpacker

lourenci:
When a Rails app doesn't have Sprockets installed and it compiles the assets through Webpacker, the Gem fails.

image

The reason is that the Gem is checking the presence of the pipeline with the `Rails.application.respond_to?(:assets)`, where it is always true even when there is no Sprockets in project. This pull fix that.

### DIFF
--- a/lib/i18n/js/dependencies.rb
+++ b/lib/i18n/js/dependencies.rb
@@ -20,7 +20,7 @@ module I18n
           assets_pipeline_available =
             (rails3? || rails4? || rails5?) &&
             Rails.respond_to?(:application) &&
-            Rails.application.respond_to?(:assets)
+            Rails.application.config.respond_to?(:assets)
           rails3_assets_enabled =
             rails3? &&
             assets_pipeline_available &&


PikachuEXE:
The related engine code:
```ruby
if JS::Dependencies.sprockets_rails_v2_plus?
  Rails.application.config.assets.configure do |config|
    config.register_preprocessor("application/javascript", ::I18n::JS::SprocketsExtension)
  end
elsif Rails.application.assets.respond_to?(:register_preprocessor)
  Rails.application.assets.register_preprocessor("application/javascript", ::I18n::JS::SprocketsExtension)
end
```
and code for `sprockets_rails_v2_plus?`:
```ruby
safe_gem_check("sprockets-rails", ">= 2")
```

If your project does not have `sprockets-rails` install, can you try to see what this return:
```ruby
if Gem::Specification.respond_to?(:find_by_name)
  Gem::Specification.find_by_name(*args)
elsif Gem.respond_to?(:available?)
  Gem.available?(*args)
end
rescue Gem::LoadError
  false
end
```

lourenci:
It returns the sprockets-rails:

image

PikachuEXE:
I see
sprockets is installed but disabled in config

PikachuEXE:
`3.0.4` released

PR #492, opened 2018-02-21
Allow i18n 1.x in gemspec

### LINKED ISSUE #491: I18n 1.0 released
The I18n gem just got bumped to 1.0. The only change is that Ruby 1.9 support is removed: https://github.com/svenfuchs/i18n/commits/master

It looks like the gemspec for this gem will need to be tweaked to remain compatible with the version bump.

ragesoss:
The i18n gem is now at 1.0; the only breaking change is removal of support for Ruby 1.9.

Resolves #491


ragesoss:
It looks like the travis config for testing against all the different versions of i18n lives on travis rather than in a .travis.yml file, but I ran `bundle exec appraisal rake spec` locally and the tests pass on the new 1.0 appraisal.

PikachuEXE:
Thanks

PikachuEXE:
Will release after Travis build pass on I18n 1.0 (running now)

PikachuEXE:
`3.0.5` released

PR #503, opened 2018-06-28
Fix extend method when translations have null values

wanovak:
When `js_extend: true`, the null values in I18n.translations are being transformed in objects.

`I18n.extend({}, {null_value: null})` results in an `array: {null_value: {}}`

This pull fix this behavior.


wanovak:
These failures were probably caused by the intermittent 500s that `npm` was giving for package installation.

PikachuEXE:
Released as `3.0.10`

PR #505, opened 2018-07-01
The element of Array may not be "string"

HunDunDM:
If the element of Array is not "null" or "string" (like Array or Object),
will get "TypeError" because of "message.match is not a function"

### DIFF
--- a/app/assets/javascripts/i18n.js
+++ b/app/assets/javascripts/i18n.js
@@ -602,7 +602,7 @@
       translation = this.interpolate(translation, options);
     } else if (isArray(translation)) {
       translation = translation.map(function(t) {
-        return this.interpolate(t, options);
+        return (typeof(t) === "string" ? this.interpolate(t, options) : t);
       }, this);
     } else if (isObject(translation) && isSet(options.count)) {
       translation = this.pluralize(options.count, scope, options);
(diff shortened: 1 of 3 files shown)


PikachuEXE:
Thanks for your contribution!
Can you also add/edit a test case to ensure this is actually fixed?

HunDunDM:
I used it indirectly when I used `react-native-i18n`.
v3.0.3 is OK, but v3.0.10 get error after upgrading because of the new Array logic which is added in v3.0.8.

```
else if (isArray(translation)) {
  translation = translation.map(function(t) {
    return this.interpolate(t, options);
  }, this);
}
```

I haven't seen the whole project yet. I'll try adding a test case.

PikachuEXE:
Just edit these 2 places
https://github.com/fnando/i18n-js/blob/v3.0.10/spec/js/translate.spec.js#L268
https://github.com/fnando/i18n-js/blob/v3.0.10/spec/js/translations.js#L62

HunDunDM:
I have added the test case.
It would get `TypeError: message.match is not a function` error in v3.0.10 before.

PikachuEXE:
Thanks!
Will release today or next week ;)

PR #516, opened 2018-10-13
added code fix and test case to pluralize the correct scope...

eriese:
... when translation is found among defaults.

Currently, if the given scope does not exist, but one of the defaults does, and the translation is meant to be pluralized, it breaks with a `[missing (original scope) translation]`. 

See test case

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Fixed
 
-- Nothing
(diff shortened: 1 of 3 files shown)


PikachuEXE:
Please revert the space removal in readme first

You might need an `.editorconfig` file like [what this project has](https://github.com/fnando/i18n-js/blob/master/.editorconfig) to avoid this from happening again

eriese:
reverted

PikachuEXE:
Sorry for the delay, got many email on Monday
Merging~

Issue #428, opened 2016-09-23
Undefined method 'initialized?' for I18n::Backend::ActiveRecord


karlkaebnick:
Am running Rails 4.2.5.1.  Have tried versions 3.0.0.rc11 as well as 2.1.2 of the i18n-js gem.  With both of them, when I run 'rake i18n:js:export' I get NoMethodError: undefined method `initialized?' for #I18n::Backend::ActiveRecord.

I know other people have seen this error for other i18n backends (and appear to have found solutions) - is there a known workaround for this with the ActiveRecord backend?

PikachuEXE:
It's due to 
https://github.com/fnando/i18n-js/blob/master/lib/i18n/js.rb#L153
And that method is assuming simple backend

stavskiys:
+1
I have the same problem

andreslemik:
try add this code to initializers:
```ruby
module I18n
  module JS
    def self.translations
     backend = ::I18n::Backend::Simple.new
     backend.instance_eval do
        init_translations unless initialized?
        Private::HashWithSymbolKeys.new(translations)
                                   .slice(*::I18n.available_locales)
                                   .to_h
      end
    end
  end
end
```

Keep in mind that in this case i18n-js will get translations from yaml instead database

Iuriy-Budnikov:
@andreslemik +1

PikachuEXE:
A new version `3.1.0` is released with #519 merged
Can you guys try it?

sbungartz:
Update: If you do try this out, mind the reloading-limitations documented in #521. Possible workarounds for use-cases with and without a `Chain`-Backend are also documented there.

vipera:
i18n-active_record 0.4.0 includes a patch that makes it compatible as a backend for i18n-js. I'm using an (ActiveRecord, Simple) chain backend and it works fine.

More details are in the PR https://github.com/svenfuchs/i18n-active_record/pull/97.

Issue #344, opened 2015-07-08
undefined method `initialized? for backend:chain


seaify:
@Sija @fnando @DefV 

Rails 4.2.1, use Asset Pipeline, I did the list of things
1. add gem "i18n-js", ">= 3.0.0.rc11" to gemfile and bundle install
2. add i18n,  i18n/translations to application.js
and in the document, it tells me that the default setting is working. but it do not
```ruby
chuck@chuck-MacBook-Pro:~/seaify/weishop(add_i18n_js|✚7…) % rake i18n:js:export                                                                                                                                                                  
rake aborted!
NoMethodError: undefined method `initialized?' for #<I18n::Backend::Chain:0x007fc76789bb70>
/Users/chuck/.rvm/gems/ruby-2.1.5/gems/i18n-js-3.0.0.rc11/lib/i18n/js.rb:155:in `block in translations'
/Users/chuck/.rvm/gems/ruby-2.1.5/gems/i18n-js-3.0.0.rc11/lib/i18n/js.rb:154:in `instance_eval'
/Users/chuck/.rvm/gems/ruby-2.1.5/gems/i18n-js-3.0.0.rc11/lib/i18n/js.rb:154:in `translations'
/Users/chuck/.rvm/gems/ruby-2.1.5/gems/i18n-js-3.0.0.rc11/lib/i18n/js.rb:88:in `translation_segments'
/Users/chuck/.rvm/gems/ruby-2.1.5/gems/i18n-js-3.0.0.rc11/lib/i18n/js.rb:34:in `export'
/Users/chuck/.rvm/gems/ruby-2.1.5/gems/i18n-js-3.0.0.rc11/lib/tasks/export.rake:5:in `block (3 levels) in <top (required)>'
/Users/chuck/.rvm/gems/ruby-2.1.5/bin/ruby_executable_hooks:15:in `eval'
/Users/chuck/.rvm/gems/ruby-2.1.5/bin/ruby_executable_hooks:15:in `<main>'
Tasks: TOP => i18n:js:export
(See full trace by running task with --trace)
```
Could you tell me how to solve this? or write a wiki for this situation. 
According to https://github.com/fnando/i18n-js/issues/59,  it seems that i should add a initializer file to set the default backend?

PikachuEXE:
What `i18n` backend you are using?

seaify:
@PikachuEXE 
I use https://github.com/spree-contrib/spree_i18n this gem, not the rails-i18n.

[...]

PR #519, opened 2018-10-26
Configurable backend

sbungartz:
Hi,
this PR addresses issues #428 and #344.
Only supporting the Simple backend to generate `translations.js` makes sense to me, since a dynamic backend like a database or something custom would raise questions about when to re-build the asset.

If you, like us, use a Chain backend with some custom stuff followed by the Simple backend and are fine with the custom stuff not being present in JS, then you can use the monkey-patch from #428.

This PR would allow to set `I18n::JS.backend = I18n::Backend::Simple.new` in an initializer, without having to monkey-patch the entire `I18n::JS.translate` method.

Quickly grepping through the project for `backend` revealed [this other use of it](https://github.com/fnando/i18n-js/blob/77e0eac361cf5e1f0cb02b2f18936f714aada92e/lib/i18n/js/fallback_locales.rb#L60), which might break if just monkey-patching `I18n::JS.translate`. I think this is a good reason, to allow this to be set from the outside.

Please let me know, what you think of this idea. If you like it, I would be happy to update README, and CHANGELOG and add/update specs. I'm not 100% sure how this might conflict with other features like `available_locales`.

### DIFF
--- a/lib/i18n/js.rb
+++ b/lib/i18n/js.rb
@@ -28,6 +28,15 @@ module I18n
       @config_file_path = new_path
     end
 
+    # Allow using a different backend than the one globally configured
+    def self.backend
+      @backend ||= I18n.backend
+    end
+
+    def self.backend=(alternative_backend)
+      @backend = alternative_backend
+    end
+
     # Export translations to JavaScript, considering settings
     # from configuration file
     def self.export
@@ -161,7 +170,7 @@ module I18n
 
     # Initialize and return translations
     def self.translations
-      ::I18n.backend.instance_eval do
+      self.backend.instance_eval do
         init_translations unless initialized?
         # When activesupport is absent,
         # the core extension (`#slice`) from `i18n` gem will be used instead
--- a/lib/i18n/js/fallback_locales.rb
+++ b/lib/i18n/js/fallback_locales.rb
@@ -57,7 +57,7 @@ module I18n
       #
       #   Maybe this should be fixed within I18n.
       def using_i18n_fallbacks_module?
-        I18n.backend.class.included_modules.include?(I18n::Backend::Fallbacks)
+        I18n::JS.backend.class.included_modules.include?(I18n::Backend::Fallbacks)
       end
 
       def ensure_valid_fallbacks_as_array!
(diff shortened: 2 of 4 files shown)


lib/i18n/js.rb
```diff
@@ -161,7 +170,7 @@ def self.filter(translations, scopes)
 
     # Initialize and return translations
     def self.translations
-      ::I18n.backend.instance_eval do
+      self.backend.instance_eval do
```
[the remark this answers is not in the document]
sbungartz: Hi,
thanks for your answer. 

It is *not* the goal of this PR to support arbitrary backends, just to allow using a different object than the one `I18n.backend` references, to allow having a more advanced `I18n.backend` while i18n-js just uses simple backend.

Probably 99% of users (who cannot user their `I18n.backend`) would set `I18n::JS.backend = I18n::Backend::Simple.new`. Then `init_translations unless initialized?` would run just fine.

It is simply a way to incorporate the idea behind the [Monkey-Patch](https://github.com/fnando/i18n-js/issues/428#issuecomment-278377233) in #428, while making all of 18n-js aware of the swapped-in simple backend.

I hope this makes sense?

PikachuEXE:
Released as `3.1.0`

lib/i18n/js.rb
```diff
@@ -161,7 +170,7 @@ def self.filter(translations, scopes)
 
     # Initialize and return translations
     def self.translations
-      ::I18n.backend.instance_eval do
```
sbungartz: @PikachuEXE Thanks for getting this released so quickly! I just tested it within our app to replace the monkey patch and now it does not load new translations when files are changed in development. This seems to be the case, because Rails calls `I18n.backend.reload!` for each request, so when the translations are fetched again from `I18n::JS.backend`, it still gets the old ones, since this instance has not been reloaded.

I was able to work around this in our case using:
```ruby
I18n::JS.backend = I18n.backend
I18n.backend = I18n::Backend::Chain.new(SomeCustomBackend.new, I18n.backend)
```

Since `Chain#reload!` calls `reload!` on all of its backends, the instance of `I18n::JS.backend` is also reloaded and everything works.

If one does not use a simple backend at all however, this is not an option.
Would it make sense to actually call `reload!` here if the backend is `initialized?`, since it is only called by the sprockets engine when something has changed anyways?

sbungartz: Yeah, you are probably right. I added some documentation in #521 with possible workarounds.

PR #523, opened 2018-11-12
Add docker for platform independent testing

illogikal:
Added a docker file, start up script, and updated readme for contribution.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,22 @@
+FROM ruby:2.5.3-alpine3.8
+ENV RAILS_ENV=development
+
+RUN apk add --update --no-cache \
+    build-base \
+    tzdata \
+    git
+
+RUN apk update && apk upgrade && \
+    apk add --no-cache bash git openssh
+
+RUN mkdir /usr/app
+WORKDIR /usr/app
+
+COPY . /usr/app/
+# COPY Gemfile.lock /usr/app/
+RUN bundle install
+
+RUN cat Gemfile.lock
+
+ADD . /usr/app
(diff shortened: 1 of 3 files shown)


PikachuEXE:
That's why we have Travis CI setup isn't it?
What cannot be archived without docker?

README.md
```diff
@@ -867,6 +867,12 @@ end
 
 See issue [#428](https://github.com/fnando/i18n-js/issues/428) for more details and discussion of this issue.
 
+## Development
```
PikachuEXE:
```suggestion
## Development with Docker only
```

Also how do you handle `bundle update`? (Since it's the layer is cached)

Dockerfile
```diff
@@ -0,0 +1,22 @@
+FROM ruby:2.5.3-alpine3.8
+ENV RAILS_ENV=development
+
+RUN apk add --update --no-cache \
+    build-base \
+    tzdata \
+    git
+
+RUN apk update && apk upgrade && \
```
PikachuEXE: Got `git` in both steps?
Can this and the `RUN` above be merged?

Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+RUN mkdir /usr/app
+WORKDIR /usr/app
+
+COPY . /usr/app/
+# COPY Gemfile.lock /usr/app/
+RUN bundle install
+
+RUN cat Gemfile.lock
+
+ADD . /usr/app
```
PikachuEXE: `ADD . /usr/app` after `COPY . /usr/app/`?

PR #397, opened 2016-03-12
Make new "I18n.extend" configurable

johnnyshields:
Many users (myself included) have issues with the new `I18n.extend` which was recently added to the Javascript.

This PR adds "js_extend" option which controls whether or not to use "I18n.extend" in your output files. This can be specified at both segment (single file) and parent level (all files).

I've also made "sort_translation_keys" to be settable at both segment and parent level to be consistent.

This PR contains some light refactoring within the `Segment` class for clarity/readability.

PR #524, opened 2018-11-12
Add json only feature for file exports/nodejs imports

illogikal:
New config key: 
`json_only default: false`

```yaml
#config/i18n-js.yml
fallbacks: false
json_only: true

translations:
  - file: "tmp/i18n-js/%{locale}.js"
```

outputs:
```
{"en":{"test":"Test"}}
```

instead of:
```
I18n.translations[\"en\"] = {"en":{"test":"Test"}};
```

This is so I can manually host the translation files behind a custom controller behind authentication using rails as an api only backend, but would be useful for anyone looking to host the files separately in an automated way.

### DIFF
--- a/lib/i18n/js.rb
+++ b/lib/i18n/js.rb
@@ -187,6 +187,13 @@ module I18n
       fallbacks != false
     end
 
+    def self.json_only
+      config.fetch(:json_only) do
+        # default value
+        false
+      end
+    end
+
     def self.fallbacks
       config.fetch(:fallbacks) do
         # default value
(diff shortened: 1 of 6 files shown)


PikachuEXE:
Thanks for your PR

There are several options like `namespace` & `pretty_print` for each segment (for each entry under `translations`) (which is processed in [`/lib/i18n/js/segment.rb`](https://github.com/fnando/i18n-js/blob/master/lib/i18n/js/segment.rb))
Example PR: #397 

Do you think it would be more flexible to add the new option there?
e.g. Generate both JSON and JS version, I dunno why someone would do that though

illogikal:
My purpose with this is to have one source of truth for webpage copy, ie. rails hosting the files and a react server importing them with no preprocessing needed.

Current options combined with json_only:
Fallbacks: can be used
Namespace: ignored
Pretty Print: can be used

I can't think someone would ever want to use both on the same application, and if they did, it's not too much work to add a PR :) If you consider it a showstopper if I don't add such functionality I can though.

PikachuEXE:
Considering `fallbacks` is also a `global` option I think it's fine (yup I missed that when commenting previously)
Let me check the code once more before merging

PR #530, opened 2019-01-17
Loosen gem dependency of i18n gem

johnnyshields:
The existing code prevents using this gem on Rails 3. The CVE itself has suitable workarounds and users themselves should be responsible for managing these. It's not the job of a secondary gem like I18n JS to protect users from it.

### DIFF
--- a/i18n-js.gemspec
+++ b/i18n-js.gemspec
@@ -18,9 +18,7 @@ Gem::Specification.new do |s|
   s.executables   = `git ls-files -- bin/*`.split("\n").map{ |f| File.basename(f) }
   s.require_paths = ["lib"]
 
-  # CVE-2014-10077 which is fixed for >= 0.8.0
-  # https://nvd.nist.gov/vuln/detail/CVE-2014-10077
-  s.add_dependency "i18n", ">= 0.8.0", "< 2"
+  s.add_dependency "i18n"
 
   s.add_development_dependency "appraisal", "~> 2.0"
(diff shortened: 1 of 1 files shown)


johnnyshields:
@PikachuEXE please merge this.

i18n-js.gemspec
```diff
@@ -18,9 +18,7 @@ Gem::Specification.new do |s|
   s.executables   = `git ls-files -- bin/*`.split("\n").map{ |f| File.basename(f) }
   s.require_paths = ["lib"]
 
-  # CVE-2014-10077 which is fixed for >= 0.8.0
-  # https://nvd.nist.gov/vuln/detail/CVE-2014-10077
-  s.add_dependency "i18n", ">= 0.8.0", "< 2"
+  s.add_dependency "i18n"
```
PikachuEXE: This does not just rollback the i18n requirement change
Before https://github.com/fnando/i18n-js/commit/b0d917acf605ba5c1615bde38ebbe1ff1bb3f12b#diff-03ae15123ab415633ccea502d7131c4a it was requiring `>= 0.6.6`

.travis.yml
```diff
@@ -8,18 +8,19 @@ cache:
   directories:
   - node_modules
 rvm:
+  - 2.1
```
PikachuEXE: Any reason to test against EOL rubies?
We *allow* usage with EOL rubies but not supporting it

johnnyshields:
@PikachuEXE both points fixed, please merge

PR #537, opened 2019-06-04
Adds strftime support for %P, %Z, and %l formats to match Ruby

dancingshell:
We have an app that currently uses Ruby `strftime` https://apidock.com/ruby/DateTime/strftime. In order to use `I18n-js` for our javascript `strftime` formatter we need to add a few missing string formats.

- `%l` (supported by Ruby), currently maps to `%-I` here. Now supports both.
- `%Z` (supported by Ruby), currently maps to `%z` here. Now supports both.
- `%P` (supported by Ruby), currently maps to `%p` here, but lower cased. Now supports `%p` and `%P`

Added tests and documentation. Would love to see these features incorporated into the library. Would appreciate comments/reviews to help get it there. Thank you!


PikachuEXE:
The doc and tests looks good, thanks!
The only thing I don't quite understand is

> `%P` (supported by Ruby), currently maps to `%p` here, but should be lower cased. I believe in the Ruby docs lowercase `%p` should result in uppercased `AM/PM` but changing that here would be a breaking change. I left `%p` as is and only added support for `%P`.

If you mean the spec is using lower-cased value like `pm` as expected target
That's due to the translation (custom?) given
https://github.com/fnando/i18n-js/blob/master/spec/js/translations.js#L85

So if the users provide upper case value, then `%P` and `%p` will work like ruby.

If I misunderstood let me know why :P

dancingshell:
@PikachuEXE that is great thank you for explaining. I didn't see the custom meridians. 👍

PikachuEXE:
Released as `3.3.0`
Also added badge with link for npm package
image

PR #546, opened 2019-09-10
add basic support for duration to strftime(), toTime()

dudekaa:
Hi

first of all, thanks for this library, even so I'm no Ruby developer, just plain JS. Thx much anyway :)

I encountered usage that hasn't been covered yet, formatting of time intervals, no matter if positive or negative. There was always client's local timezone in play, so it was kind of...inaccurate... This patch presents optional `durationFlag` which prints `Date` objects is as, w/o timezone. Maybe it can be useful for others too.

BR
Arnost

P.S. because I can't find any code style guides or such, I have done usual "docs" and "tests" stuff

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 
-- Nothing
+- optional "durationFlag" to strftime(), toTime() functions
 
 ### Changed
 
--- a/README.md
+++ b/README.md
@@ -592,6 +592,14 @@ I18n.toTime("date.formats.short", date);
 I18n.strftime(date, "%d/%m/%Y");
 ```
 
+We have some basic support for formatting durations too.
+
+```javascript
+var duration = new Date(60000); // 1 minute
+I18n.strftime("date.formats.durationHMM", duration, true);
+I18n.strftime(duration, "%-H:%M", true);
+```
+
 The accepted formats for `I18n.strftime` are:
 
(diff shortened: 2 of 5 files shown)


PikachuEXE:
I never encountered the concept `duration` in JavaScript
Do you have any reference explaining the concept and usage/examples?

dudekaa:
Of course, our particular usage is formatting duration ( length ) of movies, TV shows, etc...basically when one gets from REST API length of something in milliseconds, there was no way to format this properly using i18n-js. The word `duration` is borrowed from [Moment.js](https://momentjs.com/docs/#/durations/), so as inspiration for this feature.

PikachuEXE:
I just take a look at [moment.js doc](https://momentjs.com/docs/#/durations/) about duration
And I think it would be better to have a separate method to support duration translation as
there are multiple ways to represents a "duration"

Let me think about what to do next week
I will open another PR, just wait :)

PikachuEXE:
After some researching I think it would be faster to go for existing solutions like
- https://www.npmjs.com/package/moment-duration-format
- https://www.npmjs.com/package/duration

#### Rounding
30 days is about 1 month, but depends on context
For Feb it's > 1 month, for Jan it's < 1 month

#### Multiple form of presentation
Duration like `1 hour + 0 minutes + 50 seconds`
can be presented as (short form for easier reading)
- `1h:0m:50s`
- `1h:50s`
- `60m:50s`
- `3650s` (rare but who knows)

We can discuss more about how to solve your issue 
but I don't think duration translation would be handled by this library

PR #547, opened 2019-09-11
fix: available_locales to come from backend instead of I18n

blazeblazeblaze:
This resolves an issue, in which available_locales for fallbacks are different than the ones provided in the custom backend. The change will allow the existing behaviour (with SimpleBackend) to remain the same but fixes the value returned when the custom backend is being used.

Co-Authored-By: @jfhinchcliffe

### DIFF
--- a/lib/i18n/js.rb
+++ b/lib/i18n/js.rb
@@ -78,7 +78,7 @@ module I18n
 
     # deep_merge! given result with result for fallback locale
     def self.merge_with_fallbacks!(result)
-      I18n.available_locales.each do |locale|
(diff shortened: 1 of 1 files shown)


PikachuEXE:
Thanks for submitting a PR
Can you explain more about the issue?
What's the difference between `I18n.available_locales` and `I18n.backend.available_locales`?
Is it possible to add test case? Or can you explain how you verify this change fixes the issue?

blazeblazeblaze:
Oopsie! We've prematurely opened a pull request. We will follow up with the updated test suite and examples early next week.

PR #549, opened 2019-09-16
Add ability to import libraries on top of the translations file in an es6 way

mcamara:
In order to use rails with webpack and the [i18n-js npm package](https://www.npmjs.com/package/i18n-js) all together, `import I18n from 'i18n-js';` should be added to the resultant `translations.js` file.

This PR allows an user to add the import programatically just adding the option in the `i18n-js.yml` file. 

Thanks @PikachuEXE for your package, I hope you find this option useful 😄

### DIFF
--- a/README.md
+++ b/README.md
@@ -247,17 +247,19 @@ MyNamespace.translations || (MyNamespace.translations = {});
 MyNamespace.translations["en"] = { ... }
 ```
 
-### Import I18n library in the translations file
 
-Setting the `import: 'i18n-js'` option will add `import I18n from 'i18n-js';`. The name of the package to import can be changed.
+### Adding a line at the beggining of the translations file (useful for imports)
+
+Setting the `prefix: "import I18n from 'i18n-js';\n"` option will add the line at the beggining of the resultant translation file.
 This can be useful to use this gem with the [i18n-js](https://www.npmjs.com/package/i18n-js) npm package, which is quite useful to use it with webpack.
+The user should provide the semi-colon and the newline character if needed.
 
 For example:
 
 ```yaml
 translations:
 - file: "public/javascripts/i18n/translations.js"
-  import: 'i18n-js'
+  prefix: "import I18n from 'i18n-js';\n"
 ```
 
 will create:
@@ -265,7 +267,7 @@ will create:
 ```
 import I18n from 'i18n-js';
 I18n.translations || (I18n.translations = {});
-```
+
 
 #### Pretty Print
 
@@ -301,7 +303,8 @@ by hand or using your favorite programming language. More info below.
 #### Via NPM with webpack and CommonJS
 
 
-Add the following line to your package.json dependencies where version is the version you want
+Add the following line to your package.json dependencies  
+where version is the version you want  
 ```javascript
 "i18n-js": "{version_constraint}"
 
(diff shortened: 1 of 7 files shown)


PikachuEXE:
I currently use https://github.com/usabilityhub/rails-erb-loader:
```erb
// i18n-js.js.erb or i18n-js-ts.erb
<%# encoding: UTF-8 %>

// Optional dependency watching
// Tell webpack to watch these files
// https://github.com/usabilityhub/rails-erb-loader#dependencies
/* rails-erb-loader-dependencies ./../config/locales/ */

import I18n from "i18n-js"

I18n.translations = <%= I18n::JS.filtered_translations.to_json %>;

export default I18n
```

My erb loader file:
```js
module.exports = {
  test:     /\.erb$/,
  enforce:  "pre",
  exclude:  /node_modules/,

  use: [{
    loader:   "rails-erb-loader",
    options:  {
      runner:     (/^win/.test(process.platform) ? "ruby " : "") + "bin/rails runner",
      env:        {
        ...process.env,
        DISABLE_SPRING: 1,
        RAILS_CONFIG_EAGER_LOAD_ENABLED_IN_DEVELOPMENT: "false",
      },
    },
  }],
}
```

mcamara:
I really like your solution @PikachuEXE , but you need an extra package and change the options in webpacker, what do you think about the solution I offer in this PR? The new option in the yml file is completely optional.

README.md
```diff
@@ -282,8 +301,7 @@ by hand or using your favorite programming language. More info below.
 #### Via NPM with webpack and CommonJS
 
 
-Add the following line to your package.json dependencies
```
PikachuEXE: I think these spaces changes should be reverted

lib/i18n/js/formatters/js.rb
```diff
@@ -15,7 +15,8 @@ def format(translations)
         protected
 
         def header
-          %(#{@namespace}.translations || (#{@namespace}.translations = {});\n)
+          text = @import ? %(import I18n from '#{@import}';\n) : ''
```
PikachuEXE: I would suggest to have a `prefix` option instead of import option
So that if someone wants to add more import statements / other JS code, they can add it as well

mcamara: so the yml file would be something like:
```yml
import: "import I18n from 'i18n-js'"

```
right?

PikachuEXE: Yup
no new line required
User is responsible to add newline char / semi-colon

rafaell-lycan:
LGTM

PikachuEXE:
Still pending change from @mcamara though

mcamara:
@PikachuEXE done :)

PikachuEXE:
Thanks
Releasing next week if no other urgent pending changes

PikachuEXE:
Released in `3.4.0`

README.md
```diff
@@ -248,6 +248,27 @@ MyNamespace.translations["en"] = { ... }
 ```
 
 
+### Adding a line at the beggining of the translations file (useful for imports)
+
+Setting the `prefix: "import I18n from 'i18n-js';\n"` option will add the line at the beggining of the resultant translation file.
```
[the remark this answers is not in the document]
PikachuEXE: Fixed in https://github.com/fnando/i18n-js/commit/64d8c5a050e51078648d3b9f6bbd18b1aaa5b886

PR #554, opened 2019-11-08
added strftime format %k

JochenLutz:
We need %k format (24-hour clock, no padding) in our app to use the same formats in Rails code and in JS. It's just an alias for the existing %-H directive.

Doc and test are included.

Thanks for your work!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 
-- Nothing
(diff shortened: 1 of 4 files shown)


PikachuEXE:
Makes me wonder why it wasn't added along with `%l`...
Anyway great job and thanks!

PikachuEXE:
Released in `3.5.0`
I did `npm publish` but the npm page cache is not updated yet
But I did confirm `3.5.0` is on `npm` via `npm view i18n-js versions`

PR #560, opened 2019-12-20
Bound shortcut functions

### LINKED ISSUE #559: I18n.t is not bound
It's pretty much standard to use a shortcut for `translate` like `t('foo')`. However, it isn't bound to the context in the library.

This works:
```js
I18n.t('foo')
```

This causes an error:
```js
const {t} = I18n
t('foo')
```
The suggestion is to bind shortcuts to allow this:
```js
  I18n.t = I18n.translate.bind(I18n);
  I18n.l = I18n.localize.bind(I18n);
  I18n.p = I18n.pluralize.bind(I18n);
```
Original methods could be left intact because this benefits extendability.

bisubus:
Closes #559

### DIFF
--- a/app/assets/javascripts/i18n.js
+++ b/app/assets/javascripts/i18n.js
@@ -1084,9 +1084,9 @@
   };
 
   // Set aliases, so we can save some typing.
-  I18n.t = I18n.translate;
-  I18n.l = I18n.localize;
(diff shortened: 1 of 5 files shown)


PikachuEXE:
Merging tomorrow, sleeping in 1 hour z_z
I got my laptop with me so I think I can release it

Issue #520, opened 2018-10-31
How to load translations with requirejs (AMD)


EddieOne:
I've got something like below. Issue is that I want to load i18n-js like a module to isolate it from the page loading a SDK. However, the translation.js expects `window.I18n`. How do I load translations in this case? I'm not clear on the useage example for AMD.

```
define(function(require) {
  var i18n = require('i18n'); // i18n-js
  var translations = require('translations.js'); // js from rake i18n:js:export
...
```

PikachuEXE:
What is `translations.js`?
This one?
`https://github.com/fnando/i18n-js/blob/master/app/assets/javascripts/i18n/translations.js`

If not how do you get/generate it & what's inside?

EddieOne:
Hey, I was testing with namespaces, so it looks strange right now. I didn't find a way to make it work though. 

translations.js looks like
```
window.sdki18n.translations || (window.sdki18n.translations = {});
window.sdki18n.translations["en"] = {"sdk":{"default_table_title":"High Scores","playtomic":{"allti
...
```

I was thinking I need to customize the translations file and somehow connect it the i18n module. 

I also tried the AMD instructions, I couldn't get that to work.

EddieOne:
ah ok, I will have to make something if I want to use it this way I bet. Let's close for now

EddieOne:
@PikachuEXE I'm struggling with this still. Is it possible to add a json option per file?

```
translations:
  - file: 'sdk/app/utils/translations.js'
    only: '*.sdk.*'
    json_only: true
```

I'm thinking that I could manually set I18n.translations if I can manage to load the data.

Another option, could be add postfix, so I can get the format needed. like so

```
define(function(define) {
I18n.translations || (I18n.translations = {});
I18n.translations["en"] = {}
return I18n
}
```

PikachuEXE:
I guess you mean suffix
PR for prefix: https://github.com/fnando/i18n-js/pull/549
You can follow it and make the change and see if it solves your issue
Please test the change locally before submitting the change in a PR :)

EddieOne:
Yeah, suffix. Thanks I will try it

EddieOne:
If anyone finds this, here's the config that worked for me.

```
translations:
  - file: 'sdk/app/utils/translations.js'
    only: '*.sdk.*'
    prefix: "define(function(define) {\nI18n = {};\n"
    suffix: "return I18n;\n});"
    js_extend: false
```

Then load inside RequireJs like so

[...]

PR #561, opened 2020-02-13
Add suffix option

EddieOne:
https://github.com/fnando/i18n-js/issues/520


lib/i18n/js/formatters/js.rb
```diff
@@ -9,7 +9,7 @@ def format(translations)
           translations.each do |locale, translations_for_locale|
             contents << line(locale, format_json(translations_for_locale))
           end
-          contents
+          contents << footer
```
PikachuEXE: No point adding a method with one line only
Just use `contents << @suffix || ''`

package-lock.json
```diff
@@ -0,0 +1,192 @@
+{
```
PikachuEXE: Can we remove this file?

package.json
```diff
@@ -8,7 +8,7 @@
     "i18n"
   ],
   "devDependencies": {
-    "jasmine-node": "^1.14.5"
+    "jasmine-node": "^1.16.2"
```
PikachuEXE: Is this necessary for this PR?
I am happy to update this on `master` with enough reason
But it doesn't seem to be related to the PR

PR #564, opened 2020-03-28
Allow callback for missing translations

### LINKED ISSUE #477: Add callback when fallback is used?
Hi,

I have an app whose default language is french, and we are in progress of translating it to english.

We'd like to use fallback feature, yet be able to report and store somewhere missing english translations.

When `en -> fr` fallback is used, it seems currently not possible to run some arbitrary code to report that fallback was used (which actually means there's a missing translation).

Can we have a hook to like `missingTranslation` that notify us when fallback is used? Something like :

```javascript
I18n.onFallback = (locale, fallbackLocale, key, options) => {...}
```

geoffharcourt:
Rails makes it possible to raise an exception for a missing translation,
which can be useful in development and test environments to spot
undefined translations. This change adds an optional user-definable callback
that can be invoked any time a translation is missing for translations called in JavaScript. It could be used in tests to identify undefined translations or in production environments to report missing translations to an exception tracking
service.

When defined, the missing translation callback does not replace the
existing guess/missing translation options, it's invoked in addition to
the existing placeholders.

Close #477.

### DIFF
--- a/app/assets/javascripts/i18n.js
+++ b/app/assets/javascripts/i18n.js
@@ -189,6 +189,9 @@
     // Set missing translation behavior. 'message' will display a message
     // that the translation is missing, 'guess' will try to guess the string
     , missingBehaviour: 'message'
+    // Set a callback to be called with translations are missing. Can be used to
+    // report missing translations to exception tracking
+    , missingCallback: undefined
     // if you use missingBehaviour with 'message', but want to know that the
     // string is actually missing for testing purposes, you can prefix the
     // guessed string by setting the value here. By default, no prefix!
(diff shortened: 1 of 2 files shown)


PikachuEXE:
I think you can just assign a new `I18n.missingTranslation` function as mentioned in this comment:
https://github.com/fnando/i18n-js/issues/477#issuecomment-411509345

geoffharcourt:
I'm happy to do that, but it requires digging in to what I'm not sure are private functions like `getFullScope` to get the full key to pass back for an exception report.

PikachuEXE:
I see what you need now
I suggest passing `fullScope` and `fullScopeWithLocale` in an object to `missingCallback`
Easier to add more output to that callback if other users need something else in this callback

geoffharcourt:
I managed to make this a little easier to override by updating the type definitions, so I think I can close this out.

PR #570, opened 2020-05-28
allow options for localization and add spec

### LINKED ISSUE #568: Date format for specific locale
Maybe I don't understand this lib very well, but it seems like there's only a global locale.

What I want to do is call translate and localize using a locale. This already works for translate, I can just say:

`i18n.t('some.string', { locale: 'nl' })`

But I can't find the equivalent for localize. I'm using this library server side and so using a global locale will cause conflicts unless I set the locale before every call to localize.

Is there something I can do to force a locale for these calls? I tried using `{ locale: 'nl' }` as the second argument but it's not working.

RWOverdijk:
Fixes #568

### DIFF
--- a/app/assets/javascripts/i18n.js
+++ b/app/assets/javascripts/i18n.js
@@ -779,8 +779,8 @@
   I18n.toCurrency = function(number, options) {
     options = this.prepareOptions(
         options
-      , this.lookup("number.currency.format")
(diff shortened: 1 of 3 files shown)


RWOverdijk:
I'm not sure why Travis isn't triggered

PikachuEXE:
Release in `3.7.0`

PR #577, opened 2020-07-08
Uppercase first letter for missing translation

klautu:
'guess' is just changing underscores to spaces. Is it good idea to uppercase first letter for key?
Example:
   key - 'empty_data_set'
   missing guess - 'Empty data set'
It's just gets first letter of key, uppercase it and than join all letters except first :)


PikachuEXE:
Since there are many people using translation entry with words
I don't think it should upper case the first letter by default
You can use library like `lodash` with [`capitalize`](https://lodash.com/docs#capitalize)
If you need all guessed values to be capitalized, please assign custom `missingTranslation` function

PR #582, opened 2020-09-15
Add note for using yarn if using webpacker

mlodato517:
**This PR**
Adds a note to the README to `yarn add i18n-js` if using webpacker

**Why?**
I'm not actually sure if this is the right way to handle things but I followed the README for "without asset pipeline" and then got stuck because there was no `I18n` on the client. I was pointed to [this gist](https://gist.github.com/bazzel/ecdff4718962e57c2d5569cf01d332fe) and after adding the `i18n-js` package to the client, "everything worked".

If this isn't the right advice, I'd be happy to help update it to whatever it needs to be and/or point users in the right direction if this is already documented elsewhere!


PikachuEXE:
Thanks for the PR
I think it better be moved to after `Add the gem to your Gemfile.` and before `Rails app with Asset Pipeline`

Since it's "optional step for installation"

README.md
```diff
@@ -79,6 +79,16 @@ Then get the JS files following the instructions below.
   then only the specified locales will be exported.
   Current version of `i18n.js` will also be exported to avoid version mismatching by downloading.
 
+#### Rails with [webpacker](https://github.com/rails/webpacker)
+
+If you're using `webpacker`, you may need to add the dependencies to your client with:
+
+```
+yarn add i18n-js
```
PikachuEXE, line 87:
```shell
# Depends on your package manager
npm install i18n-js
# or
yarn add i18n-js
```

PikachuEXE (changes requested):
We should cover all common package managers not just `yarn` (even though I use `yarn`)

README.md
mlodato517: Okay! I chose `yarn` since it's what `webpacker` uses and you have to go through some hoops to use `npm` with `webpacker` but this sounds good!

mlodato517: dbab13a

PikachuEXE (approved):

PikachuEXE:
Thanks for your contribution
Welcome to make more if you see fit

PR #583, opened 2020-09-30
Allow scoped calls to toHumanSize()

scumah:
Hey! This PR adds support for the `scope` option to the `toHumanSize()` function.

Right now, the function directly looks for a specific hardcoded scope that doesn't take into account if the user is passing a `scope` option. In our case, we wanted to have this kind of translations under a `shared` scope, along the date formats and that kind of stuff, but using `toHumanSize()` in it's current state makes it impossible.

To fix it we calculate the scope before calling `t()` using `getFullScope()`, passing the options `toHumanSize()` got.

Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -337,7 +337,7 @@ var i18n = require("i18n-js");
 
 ### Setting up
 
-You **don't** need to set up a thing. The default settings will work just okay. But if you want to split translations into several files or specify specific contexts, you can follow the rest of this setting up section.
+You **don't** need to set up a thing. The default settings will work just okay. But if you want to split translations into several files or specify contexts, you can follow the rest of this setting up section.
 
 Set your locale is easy as
 ```javascript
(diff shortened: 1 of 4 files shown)


PikachuEXE:
Sorry I am on holiday
Come back next week

PikachuEXE:
That seems a great addition
Can you also update the README to reflect your changes?

scumah:
Done @PikachuEXE. I took the chance and:
- Fixed a couple of typos
- Updated the plural rules broken link
- Removed trailing white spaces

PikachuEXE:
Hey thx
But I notice some trailing double spaces are removed which is for formatting
Can you add back those double spaces to keep the formatting right? (Unless that's also part of the fix)

scumah:
wops! I didn't know about this double space for md, sry. fixing it!

PR #591, opened 2021-01-08
! Fix proc exported to JS/JSON file(s) causing issues like git merge conflicts

### LINKED ISSUE #439: Procs are exported as their :to_s representation
When exporting translations to js, the resulting JSON contains following:
`"rule": "#<Proc:0x0074823784723@/.rvm/gems/ruby-2.3.1/gems/rails-i18n-5.0.1/lib/rails_i18n/common_pluralizations/one_other.rb:7 (lambda)>"`
under `i18n.plural.rule`. This entry changes every time the translations.js file is regenerated, which causes git conflicts.

Since there is no way we can convert the procs to js, would the best course of action be to just remove all procs from the translations before exporting them?

PikachuEXE:
Fixes #439

### DIFF
--- a/lib/i18n/js/segment.rb
+++ b/lib/i18n/js/segment.rb
@@ -50,6 +50,7 @@ module I18n
       def write_file(_file = @file, _translations = @translations)
         FileUtils.mkdir_p File.dirname(_file)
         _translations = Utils.deep_key_sort(_translations) if @sort_translation_keys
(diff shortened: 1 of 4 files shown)


arturo-ojeda:
God please, have this PR approved 🙏🏻

PikachuEXE:
Got spec failed after merge but probably not caused by this PR
Will release after fixing failures

Issue #592, opened 2021-01-29
Our i18n-js.yml gets parsed 240 times on first page load


Shyked:
We are using the ERB syntax to generate a configuration file reflecting the languages list we have in our database.

```
fallbacks: false

translations:
  - file: "app/assets/javascripts/i18n/base.js"
    only: "base.*"
    pretty_print: true
<% Language.all.each do |language| %>
  - file: <%= "app/assets/javascripts/i18n/#{language.code}.js" %>
    only: <%= "#{language.code}.*" %>
    pretty_print: true
<% end %>
  - file: <%= "app/assets/javascripts/i18n/en.js" %>
    only: <%= "en.*" %>
    pretty_print: true
```

We noticed that this configuration is causing 240 calls to our database coming from i18n-js.

In `/lib/i18n/js.rb`, I see the `config` function is called multiple times, and everytime it performs a new `ERB.new(File.read(config_file_path)).result`.

Is there a way to avoid this behavior, maybe a workaround, or is there something we're not doing correctly?

We're on i18n-js 3.3.0, I didn't try the latest version yet but after looking at the `js.rb` file I assumed the behavior would stay the same as the config is not cached.

PikachuEXE:
This is due to your config file having
`Language.all.each`
If that is a ORM class (e.g. `ActiveRecord::Base`)
Can you provide more details about it?
How's it structured, what's stored, how frequently it's updated...

Shyked:
Thank you for your response.
It is updated only during migrations so our database reflects the languages we support.

```
#<Language:0x00005654bf4aeba0
 id: 23,
 name: "Afrikaans",
 code: "af",
 created_at: Mon, 14 Dec 2020 05:04:37 EST -05:00,
 updated_at: Mon, 14 Dec 2020 05:04:37 EST -05:00,
 country_code: nil,
 name_in_locale: nil>
```

Here's a sample of our server logs

```
Started GET "/users/sign_in" for 127.0.0.1 at 2021-02-01 09:42:25 +0100
Processing by SessionsController#new as HTML
  Rendering devise/sessions/new.html.slim within layouts/application
  Rendered shared/_widget_code.html.erb (2.1ms)
  Rendered shared/_background_signin.html.slim (45.5ms)
  Language Load (0.5ms)  SELECT `languages`.* FROM `languages` ORDER BY `languages`.`name_in_locale` ASC, `languages`.`name` ASC
  ↳ (erb):35
[...]

PR #593, opened 2021-02-05
* Update `I18n::JS.config` to read config from file with caching

PikachuEXE:
Config object cached until new config file path assigned
This solves performance issue for config files with heavy lifting (e.g. DB query)

Performance issue report:
https://github.com/fnando/i18n-js/issues/592

### DIFF
--- a/lib/i18n/js/private/config_store.rb
+++ b/lib/i18n/js/private/config_store.rb
@@ -1,0 +1,31 @@
+# frozen_string_literal: true
+
+require "singleton"
+
+module I18n
+  module JS
+    # @api private
+    module Private
+      # Caching implementation for I18n::JS.config
+      #
+      # @api private
+      class ConfigStore
+        include Singleton
+
+        def fetch
+          return @config if @config
+
+          yield.tap do |obj|
+            raise AugumentError, "unexpected falsy object from block" unless obj
+
+            @config = obj
+          end
+        end
+
+        def flush_cache
+          @config = nil
+        end
+      end
+    end
+  end
+end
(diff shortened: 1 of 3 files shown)


Shyked:
I've put two DB queries in my i18n-js.yml file (for testing purposes), called `I18n::JS.filtered_translations.to_json`, and ended up with also two DB queries logged, this PR solves my issue 👌 

```Ruby
[2] pry(main)> I18n::JS.filtered_translations.to_json
  # Language Load (0.2ms)  SELECT `languages`.* FROM `languages` ORDER BY `languages`.`name_in_locale` ASC, `languages`.`name` ASC
  # Language Load (0.2ms)  SELECT `languages`.* FROM `languages` ORDER BY `languages`.`name_in_locale` ASC, `languages`.`name` ASC
=> "{\"af\":{\"activerecord\":{\"attributes\":...
```

PikachuEXE:
Released in `3.8.1`

lib/i18n/js/private/config_store.rb
```diff
@@ -0,0 +1,31 @@
...
+      #
+      # @api private
+      class ConfigStore
+        include Singleton
+
+        def fetch
+          return @config if @config
+
+          yield.tap do |obj|
+            raise AugumentError, "unexpected falsy object from block" unless obj
```
kostassof, line 19: Hello. Is there an `AugumentError` exception?

PikachuEXE: There is:
https://ruby-doc.org/core-2.7.0/ArgumentError.html

kostassof: I'm aware but there is a typo here. :)

PikachuEXE: Arrr
Let me fix

kostassof: Thanks.

PR #598, opened 2021-03-11
Fix deprecation warning

hasghari:
Fixes the following warning:

```
warning: File.exists? is deprecated; use File.exist? instead
```


PikachuEXE:
Just search the whole repo and confirmed this is the only `File.exists?` left
Thanks for spotting & the PR!

PikachuEXE:
Released in `3.8.2` 👋

PR #605, opened 2021-05-17
JSON.parse approach is much faster than object literal

pastak:
## `JSON.parse` is much faster than object literal

Because the JSON grammar is much simpler than JS's object literal so make it faster about x1.5 on some browser's JS engines, not only V8 with Chrome but also JavaScriptCore with Safari or Chakra with Internet Explorer

cf. https://v8.dev/blog/cost-of-javascript-2019#json

## with `pretty_print` option

Pretty-printed JSON should be contained into template string  to pass `JSON.parse` but  Internet Explorer 11 doesn't support template string so it avoid using `JSON.parse` when `pretty_print` option is `true`.

### DIFF
--- a/lib/i18n/js/formatters/js.rb
+++ b/lib/i18n/js/formatters/js.rb
@@ -20,11 +20,11 @@ module I18n
         end
 
         def line(locale, translations)
-          json_stringified = @pretty_print ? %(`#{translations}`) : %('#{translations}')
+          json_literal = @pretty_print ? translations : %(JSON.parse('#{translations}'))
           if @js_extend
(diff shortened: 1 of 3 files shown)


PikachuEXE:
Thanks for providing link to the article
As the tests are passing I will just release this soon and let people test it (maybe they won't notice a thing)

pastak:
@PikachuEXE Oh, I find a problem with JSON contains `'`...
```

PikachuEXE:
Try to create a failing case
We also have an option called `json_only` (not well documented)
Introduced in https://github.com/fnando/i18n-js/pull/524
I will hold my release until you figure if single quotes would cause issue after this PR merged

pastak:
I have became aware `json_only` and it keeps same results as before 👍

PR #607, opened 2021-05-17
escape double quote and backslash

pastak:
Bugfix: https://github.com/fnando/i18n-js/pull/605

I tried new feature with `JSON.parse` in our Rails application and find bug to fix 🙇 

- `"` is escaped  as `\"` but it should escape as  `\\"`
  - `\` should escape as `\\`

I added test escaping on Ruby and parsing on JavaScript.

### DIFF
--- a/lib/i18n/js/formatters/js.rb
+++ b/lib/i18n/js/formatters/js.rb
@@ -2,7 +2,8 @@ require "i18n/js/formatters/base"
 
 JSON_ESCAPE_MAP = {
   '\\"' => '\\\\"',
-  "'" => "\\'"
+  "'" => "\\'",
+  "\\" => "\\\\"
 }
 
 module I18n
@@ -25,7 +26,7 @@ module I18n
         end
 
         def line(locale, translations)
-          json_literal = @pretty_print ? translations : %(JSON.parse('#{translations.gsub(/\\"|'/){|match| JSON_ESCAPE_MAP[match] }}'))
+          json_literal = @pretty_print ? translations : %(JSON.parse('#{translations.gsub(/\\"|'|\\/){|match| JSON_ESCAPE_MAP[match] }}'))
           if @js_extend
             %(#{@namespace}.translations["#{locale}"] = I18n.extend((#{@namespace}.translations["#{locale}"] || {}), #{json_literal});\n)
           else
--- a/lib/i18n/js/formatters/js.rb
+++ b/lib/i18n/js/formatters/js.rb
@@ -1,5 +1,10 @@
 require "i18n/js/formatters/base"
 
+JSON_ESCAPE_MAP = {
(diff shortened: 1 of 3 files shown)


pastak:
TODO

- [ ] check about `\n`

pastak:
note: It is no issue with `\n`

lib/i18n/js/formatters/js.rb
```diff
@@ -1,5 +1,10 @@
 require "i18n/js/formatters/base"
 
+JSON_ESCAPE_MAP = {
+  "'" => "\\'",
+  "\\" => "\\\\"
+}
```
PikachuEXE, line 6: 1. This hash should be frozen
2. This constant should be placed inside the class so that it does not affect top namespace
3. Good to have `private_constant : JSON_ESCAPE_MAP` after it's moved into class
https://docs.rubocop.org/rubocop/1.0/cops_style.html#styleconstantvisibility

lib/i18n/js/formatters/js.rb
```diff
@@ -20,7 +25,7 @@ def header
         end
 
         def line(locale, translations)
-          json_literal = @pretty_print ? translations : %(JSON.parse('#{translations.gsub("'"){"\\'"}}'))
+          json_literal = @pretty_print ? translations : %(JSON.parse('#{translations.gsub(/'|\\/){|match| JSON_ESCAPE_MAP[match] }}'))
```
PikachuEXE, line 28: Possible to use the keys of the constant hash?
Also use `Hash#fetch` to avoid returning `nil` silently
```suggestion
          json_literal = @pretty_print ? translations : %(JSON.parse('#{translations.gsub(/#{JSON_ESCAPE_MAP.keys.join("|")}/){|match| JSON_ESCAPE_MAP.fetch(match) }}'))
```

PikachuEXE (changes requested):
Probably only this 2 changes needed

lib/i18n/js/formatters/js.rb
pastak: Impossible to use `keys.join` because `too short escape sequence` error (maybe caused by unescaped key) so I use `Regexp.union`

pastak:
@PikachuEXE  I change with your suggestions 👍

PikachuEXE (approved):

PikachuEXE:
Will be released tomorrow (unless too busy)

razum2um:
@pastak @PikachuEXE even after 2 fixed I think this still doesn't handle all cases, it should be `escape_javascript` from rails, see theirs map https://api.rubyonrails.org/classes/ActionView/Helpers/JavaScriptHelper.html

PR welcome or wdyt?

PikachuEXE:
PRs are welcome as long as there are enough test cases and they all passed

Issue #615, opened 2021-07-09
How to generate translations for locales which are not in the  I18n.available_locales


avokhmin:
Hello!

> If you can't or prefer not to generate this file, you can move the middleware line to your config/environments/development.rb file and run rake i18n:js:export before deploying. This will export all translation files, including the custom scopes you may have defined on config/i18n-js.yml. If I18n.available_locales is set (e.g. in your Rails config/application.rb file) then only the specified locales will be exported. Current version of i18n.js will also be exported to avoid version mismatching by downloading.

How can I add extra locales which are not presented in the `I18n.available_locales`.

User case: we have the site in EN/JP, but in JS widgets we would like to use more locales.

Something like:

```
config/i18n-js.yml

translations:
  - file: "public/javascripts/translations.js"
    only: "*.js.*"
```

```
> I18n.available_locales
=> [:en, :jp]
> I18n::JS.backend.available_locales.first(4)
=> [:en, :cs, :tr, :"pt-BR"]
```

PikachuEXE:
I think right now it's all "hardcoded" to use `I18n.available_locales`

How do you generate/pull the translations at the moment? 
I can write some patch code for your use case first.

avokhmin:
> I think right now it's all "hardcoded" to use `I18n.available_locales`
> 
> How do you generate/pull the translations at the moment?
> I can write some patch code for your use case first.

@PikachuEXE  we use `v2.1.2` as TMP solution :)

avokhmin:
@PikachuEXE I suppose that https://github.com/fnando/i18n-js/pull/547 can help, but need to test. What do you think?

PikachuEXE:
I think #547 is not good enough

`I18n.available_locales` should be the default and we can update the gem to allow custom `available_locales`
What do you think?

avokhmin:
> and we can update the gem to allow custom `available_locales`
> What do you think?

@PikachuEXE I like this idea!

avokhmin:
@PikachuEXE do you have a plan to implement it? Or I can try to do it...

PikachuEXE:
Please submit a PR for it
Thx

PR #617, opened 2021-07-28
Allow to set available locales

avokhmin:
PR for https://github.com/fnando/i18n-js/issues/615

This PR allows to set custom locales instead of using only `I18n.available_locales`

### DIFF
--- a/README.md
+++ b/README.md
@@ -273,8 +273,10 @@ tests.
 
 #### Available locales
 
-Setting the `js_available_locales` option allows to generate translations not only
-for `I18n.available_locales` locales (which we use by default). Example:
+By specifying option `js_available_locales` with a list of locales, this list
+would be used instead of default `I18n.available_locales` to generate translations.
+
+Example:
 
 ```yaml
 js_available_locales: ["de", "en"]
--- a/README.md
+++ b/README.md
@@ -271,6 +271,15 @@ You must disable this feature by setting the option to `false`.
 To find more examples on how to use the configuration file please refer to the
 tests.
 
+#### Available locales
+
+Setting the `js_available_locales` option allows to generate translations not only
+for `I18n.available_locales` locales (which we use by default). Example:
+
+```yaml
+js_available_locales: ["de", "en"]
+```
+
 #### Namespace
 
 Setting the `namespace` option will change the namespace of the output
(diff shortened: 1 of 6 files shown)


avokhmin:
@PikachuEXE can you review PR?

lib/i18n/js.rb
```diff
@@ -183,7 +183,7 @@ def self.translations
         #
         # So the input is wrapped by our class for better `#slice`
         Private::HashWithSymbolKeys.new(translations).
-          slice(*::I18n.available_locales).
+          slice(*::I18n::JS.js_available_locales).
```
avokhmin, line 186: I think that better to use `js_available_locales` name for new method, because when I used:

```
slice(*available_locales).
```

I got some mystic errors, until I found that we use `self.backend.instance_eval do` here.

So, maybe the other name of method will help to support gem with lilttle less pain in the future :)

README.md
```diff
@@ -271,6 +271,15 @@ You must disable this feature by setting the option to `false`.
 To find more examples on how to use the configuration file please refer to the
 tests.
 
+#### Available locales
+
+Setting the `js_available_locales` option allows to generate translations not only
+for `I18n.available_locales` locales (which we use by default). Example:
```
PikachuEXE, line 277:
```suggestion
By specifying option `js_available_locales` with a list of locales, this list would be used instead of default `I18n.available_locales` to generate translations.  
Example:
```

lib/i18n/js.rb
PikachuEXE: I agree with your naming which matches its purpose.

spec/ruby/i18n/js_spec.rb
```diff
@@ -307,21 +307,17 @@
       expect(subject[:de][:null_test]).to eql(nil)
     end
 
-    context "when given locale is in `I18n.available_locales` but its translation is missing" do
```
PikachuEXE, line 310: I think the original test cases should be kept.  
And copy `I18n.available_locales` test cases to a new context for `js_available_locales` and modify those copied test cases for `js_available_locales`.

PikachuEXE (changes requested):

avokhmin:
@PikachuEXE , thanks for review! PR updated.

PikachuEXE (approved):
Briefly checked the updated spec & README
LGTM

PikachuEXE:
Released in `3.9.0`

Issue #622, opened 2021-09-03
v4 translations are removed when server starts


michelson:
### Discussed in https://github.com/fnando/i18n-js/discussions/621

<div type='discussions-op-text'>

<sup>Originally posted by **michelson** August 26, 2021</sup>
Hello, 
I've integrated the v4 in [our app](https://github.com/chaskiq/chaskiq), and we see that when the server starts (dev env), the JSON translations are emptied. 
How can we avoid this? have you seen this behavior?

thanks</div>

fnando:
Closed via #631

PR #623, opened 2021-09-05
notes on listen, after initialize

michelson:
Added readme notes on I18nJS.listen usage. The current suggestion on how add I18nJS.listen on an initializer leads to empty translations when server starts. If the I18nJS.listen is added in an after_initialize it works fine.

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,7 +96,7 @@ Now you can run `guard start -i`.
 
 ### Using listen
 
-Create a file under `config/environments/development.rb` with the following content:
+Add the following to `config/environments/development.rb` file:
 
 ```ruby
 config.after_initialize do
--- a/README.md
+++ b/README.md
@@ -96,14 +96,16 @@ Now you can run `guard start -i`.
 
 ### Using listen
 
-Create a file under `config/initializers/i18n.rb` with the following content:
+Create a file under `config/environments/development.rb` with the following content:
 
 ```ruby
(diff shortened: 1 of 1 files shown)


michelson:
fixes #622

README.md
```diff
@@ -96,14 +96,16 @@ Now you can run `guard start -i`.
 
 ### Using listen
 
-Create a file under `config/initializers/i18n.rb` with the following content:
+Create a file under `config/environments/development.rb` with the following content:
```
michelson, line 99:
```suggestion
Add the following to `config/environments/development.rb` file:
```

michelson:
I've found another problem, the listen is evaluating an already initialized version of translations so if there is any change it will not update the files. I've commented the `unless` condition on `init_translations` in order for the content be updated properly.

```ruby
  def self.translations
    ::I18n.backend.instance_eval do
      has_been_initialized_before =
        respond_to?(:initialized?, true) && initialized?
      puts "INIT BEFORE #{has_been_initialized_before.to_s}"
      init_translations #unless has_been_initialized_before <--- init_translations has to run always
      translations
    end
  end
```

fnando:
Closed via #631.

fnando:
@michelson thanks the pr. And sorry for over sighting this for so long.

As for the exporting files only once, I fixed it on 9e8d432aa1375aed34d0085f6db3926434b2ac8e.

PR #626, opened 2021-11-16
Update README with details on backend compatibility

vipera:
As i18n-active_record supports being used as a backend for i18n-js since 0.4.0 I'm opening this patch to correct the related part of the docs.


PikachuEXE:
I guess you mean https://github.com/svenfuchs/i18n-active_record/releases/tag/v0.4.0 then
I will merge it now

PR #633, opened 2022-01-13
CI against Ruby 3.0 and 3.1

mishina2228:
This PR adds Ruby 3.0 and 3.1 to the CI matrix.
I have confirmed that it is all green in my forked repo.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -26,6 +26,8 @@ jobs:
           - 2.5
           - 2.6
           - 2.7
+          - '3.0'
+          - 3.1
         gemfile:
           - gemfiles/i18n_0_6.gemfile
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yaml
```diff
@@ -26,6 +26,8 @@ jobs:
           - 2.5
           - 2.6
           - 2.7
+          - '3.0'
+          - 3.1
```
PikachuEXE, line 30: Maybe just to avoid parsing issue maybe we should quote all future versions with quotes?
```suggestion
          - '3.1'
```

mishina2228: hmm, the problem only occurs with 'x.0', so we can leave this as it is.
But either way is fine with me!

PikachuEXE:
I can change it myself later anyway

PR #645, opened 2022-03-25
Fix pluralization fallback in i18n.js

ryanb:
I ran into this error with pluralization in I18n.js v3.8.0 when fallbacks are enabled.

> Error: TypeError: null is not an object (evaluating 'translations[pluralizerKey]')

Here are the steps to reproduce:

```js
I18n.translations = {en: {items: {one: "1 item", other: "%{count} items"}}, fr: {items: null}};
I18n.defaultLocale = "en";
I18n.locale = "fr";
I18n.fallbacks = true
I18n.t("items", {count: 1})
// Error: TypeError: null is not an object (evaluating 'translations[pluralizerKey]')
```

From my understanding this gem provides fallback support in two different ways: Ruby side and JavaScript side. I have disabled the Ruby fallbacks since it clones the locale value which is inefficient for how I have it setup. I'd like to rely on the JavaScript fallbacks but ran into this error.

Checking if `translations` exists before calling `pluralizerKey` seems to fix this issue.

### DIFF
--- a/app/assets/javascripts/i18n.js
+++ b/app/assets/javascripts/i18n.js
@@ -433,7 +433,7 @@
       , pluralizerKey
       , message;
 
-    if (isObject(translations)) {
+    if (translations && isObject(translations)) {
       while (pluralizerKeys.length) {
         pluralizerKey = pluralizerKeys.shift();
         if (isSet(translations[pluralizerKey])) {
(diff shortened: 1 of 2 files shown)


PikachuEXE:
Will be released next week if no other PRs

ryanb:
Awesome, thanks for your work on this!

PikachuEXE:
Released
See [v3 branch](https://github.com/fnando/i18n-js/tree/v3) badges for released versions

PR #649, opened 2022-05-03
Add test examples for using scope option with defaults

afdev82:
# What
I've added two new test examples, following the hint from @PikachuEXE [here](https://github.com/fnando/i18n-js/discussions/648#discussioncomment-2677441).

# Why
It seems that the fallback to the default scope is not working when the scope option is present.
Instead of giving me the scope in the defaults, it returns a missing translation message.

### DIFF
--- a/spec/js/translate.spec.js
+++ b/spec/js/translate.spec.js
@@ -155,6 +155,21 @@ describe("Translate", function(){
       expect(actual).toEqual("Hello stranger!");
     });
 
+    it("uses scope provided in defaults if scope as an array doesn't exist", function() {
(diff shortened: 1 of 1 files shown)


afdev82:
Hi @fnando, yes maybe that PR was not needed.
But the test addition was done to support the discussion [here](https://github.com/fnando/i18n-js/discussions/648#discussioncomment-2677441) and possibly find a fix for that.
@PikachuEXE was asking for your opinion [here](https://github.com/fnando/i18n-js/discussions/648#discussioncomment-2678907), could you support on that? Thank you!

PR #661, opened 2022-08-18
Support multiple locales dirs in listen backend.

morgoth:
In addition allow to configure Listen by passing options.

The usecase: having multiple places from where the locales are loaded.
We're using https://github.com/tiramizoo/fractual_i18n but it can be any other generic case.

This helps us to configure backend via:
```ruby
  I18nJS.listen(
    locales_dir: [Rails.root.join("config/locales"), Rails.root.join("app/views")],
    options: {only: %r{.yml$}}
  )
```

### DIFF
--- a/lib/i18n-js/listen.rb
+++ b/lib/i18n-js/listen.rb
@@ -7,7 +7,8 @@ module I18nJS
 
   def self.listen(
     config_file: Rails.root.join("config/i18n.yml"),
-    locales_dir: Rails.root.join("config/locales")
+    locales_dir: Rails.root.join("config/locales"),
+    options: {}
   )
     return unless Rails.env.development?
     return if started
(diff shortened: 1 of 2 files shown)


lib/i18n-js/listen.rb
```diff
@@ -7,7 +7,8 @@ class << self
 
   def self.listen(
     config_file: Rails.root.join("config/i18n.yml"),
-    locales_dir: Rails.root.join("config/locales")
+    locales_dir: Rails.root.join("config/locales"),
```
morgoth, line 10: I guess this option could be now deprecated in favor of `locales_dirs` but it can be done in separate PR

morgoth:
@fnando what do you think about this PR? Can we have it merged?

fnando (approved):

PR #688, opened 2022-12-24
feat: speed up fallback translations plugin

oleksii-leonov:
<details>

### PR Structure

      PRs).
      two PRs otherwise).

### Thoroughness

      fixes.
- [x] I've updated any docs, `.md` files, etc… affected by this change.

</details>

### What

Speed up the fallback translations plugin.
`deep_merge` code extracted from `i18n-js` v3.

### Why

Current versions with `Glob` became too slow on large localization files.
For our use case (20 locales, 20000 keys each), I wait ~20 minutes and kill the process.

`deep_merge` versions from `v3` works instantly. So, I have extracted the `deep_merge` from `v3` into `v4`.

### Known limitations

N/A

### DIFF
--- a/lib/i18n-js/embed_fallback_translations_plugin.rb
+++ b/lib/i18n-js/embed_fallback_translations_plugin.rb
@@ -4,6 +4,34 @@ module I18nJS
   require "i18n-js/plugin"
 
   class EmbedFallbackTranslationsPlugin < I18nJS::Plugin
+    module Utils
+      # Based on deep_merge by Stefan Rusterholz, see <https://www.ruby-forum.com/topic/142809>.
+      # This method is used to handle I18n fallbacks. Given two equivalent path nodes in two locale trees:
+      # 1. If the node in the current locale appears to be an I18n pluralization (:one, :other, etc.),
+      #    use the node, but merge in any missing/non-nil keys from the fallback (default) locale.
+      # 2. Else if both nodes are Hashes, combine (merge) the key-value pairs of the two nodes into one,
+      #    prioritizing the current locale.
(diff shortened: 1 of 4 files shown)


fnando:
This is great. Thank you! I did just a few synthetic tests with no more than a hundred translations and it's okay, but your real use case caught the performance introduced by glob (i'm pretty sure there's an opportunity for optimization there).

lib/i18n-js/embed_fallback_translations_plugin.rb
```diff
@@ -16,33 +44,21 @@ def validate_schema
...
-        buffer[locale] = Glob.new(translations[locale]).tap do |glob|
-          glob << "*"
-        end
-      end
-
-      default_locale = I18n.default_locale
-      default_locale_glob = mapping.delete(default_locale)
-      default_locale_paths = default_locale_glob.paths
+      fallback_locale = I18n.default_locale.to_sym
+      locales_to_fallback = translations.keys - fallback_locale
```
fnando, line 50: Shouldn't this be `[fallback_locale]` instead? I'm running this in development and I'm getting an exception.

```suggestion
      locales_to_fallback = translations.keys - [fallback_locale]
```

fnando:
@oleksii-leonov just pushed to your repo with some changes related to uncovered code (just to make sure those mergers were working as expected).

lib/i18n-js/embed_fallback_translations_plugin.rb
oleksii-leonov: Sorry, my bad. It should be `translations.keys - [fallback_locale]` 👍

PR #733, opened 2025-03-24
Add `parallel` option

aergonaut:
<details>

### PR Structure

      PRs).
      two PRs otherwise).

### Thoroughness

      fixes.
- [x] I've updated any docs, `.md` files, etc… affected by this change.

</details>

### What

This PR adds a `parallel` option to `I18nJs.call`, which is passed to the `export_group` method. When `parallel: true` and the group uses `:locale` placeholder, `export_group` will use concurrent-ruby's Promise to parallelize exporting each file.

I have included a benchmark script in this PR which compares the time needed to export 100 locale files with 5_000 strings each. Here are some example results:

```
❯ ruby test/benchmarks/parallel_export.rb
              user     system      total        real
baseline  4.199457   0.144715   4.344172 (  4.376356)
parallel  1.999492   0.071311   2.070803 (  2.064344)
```

Parallelization makes the process takes about 50% less time.

### Why

In an app with many strings and many locales, exporting can take a significant amount of time, because each locale is exported sequentially. This slows down the feedback loop, making developers wait longer to see the effects of a change.

### Known limitations

[TODO or N/A]


fnando:
Closed by #748

PR #736, opened 2025-10-31
Remove benchmark gem

morgoth:
It issues a deprecation warning: benchmark was loaded from the standard library, but will no longer be part of the default gems starting from Ruby 3.5.0.

Since the used method is very simple, there is no need for dependency


fnando:
Merged this changed via #737. Thanks!

fnando:
Just released 4.2.4 with this change!

PR #740, opened 2026-03-27
Update README to document .* suffix for nested key exclusion patterns

artinboghosian:
### What

The glob gem v0.4.1 changed pattern matching to anchor non-wildcard patterns with `$`, meaning `!*.activerecord` no longer excludes nested keys like `en.activerecord.attributes`. Documents the correct pattern `!*.activerecord.*` to properly exclude all nested translations. 

https://github.com/fnando/glob/commit/7baf6cd5d49e6b3df8343da78b602249bcdece55

### Why

After upgrading the glob gem to v0.4.1, exclusion patterns like `!*.activerecord` stopped matching nested keys. The README showed the same syntax, so it wasn't clear why the documented patterns weren't working. This updates the README to use `!*.activerecord.*` so the documentation reflects what actually works with glob v0.4.1+. The updated patterns are also backwards compatible with older versions of glob.

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,9 +68,9 @@ translations:
   - file: app/frontend/locales/en.json
     patterns:
       - "*"
-      - "!*.activerecord"
-      - "!*.errors"
-      - "!*.number.nth"
+      - "!*.activerecord.*"
+      - "!*.errors.*"
+      - "!*.number.nth.*"
 
   - file: app/frontend/locales/:locale.:digest.json
     patterns:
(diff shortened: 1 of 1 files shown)


artinboghosian:
@fnando I'm aware that tests are failing, don't want you to think I'm ignoring that but they are unrelated to any changes I made. They are rubocop linting errors. I only modified README.md

artinboghosian:
@fnando bumping

artinboghosian:
@fnando bumping

artinboghosian:
@fnando bumping

artinboghosian:
@fnando thanks for merging and maintaining this library. My team and I find it extremely useful. Thanks.

PR #742, opened 2026-04-03
Fix rubocop warnings

fukayatsu:
<details>

### PR Structure

      PRs).
      two PRs otherwise).

### Thoroughness

      fixes.
- [ ] I've updated any docs, `.md` files, etc… affected by this change.

</details>

### What

Fix rubocop warnings by replacing `Class.new(StandardError)` with explicit class inheritance syntax, and replacing `each_with_object` with `to_h` for cleaner hash construction.

### Why

These changes address rubocop style warnings to keep the codebase consistent with Ruby conventions:

- `Class.new(StandardError)` triggers `Style/ClassAndModuleChildren` or similar cops. Using `class FooError < StandardError; end` is the more idiomatic and readable Ruby pattern.
- `each_with_object({})` for building a hash can be simplified with `to_h`, which is the preferred rubocop style (`Style/HashTransformKeys` / `Style/HashTransformValues`).

### Known limitations

N/A

### DIFF
--- a/lib/i18n-js.rb
+++ b/lib/i18n-js.rb
@@ -16,7 +16,7 @@ require_relative "i18n-js/sort_hash"
 require_relative "i18n-js/clean_hash"
 
 module I18nJS
-  MissingConfigError = Class.new(StandardError)
+  class MissingConfigError < StandardError; end
 
   def self.call(config_file: nil, config: nil)
     if !config_file && !config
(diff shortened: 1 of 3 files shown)


fukayatsu:
The CI workflow uses `pull_request_target` with `actions/checkout@v4` without specifying `ref`, which means it checks out the base branch (main) code rather than the PR head. As a result, RuboCop is running against the pre-fix code, causing the same offenses this PR addresses.

CI should pass once this is merged and runs against main.

fnando:
I ended up disabling those rules, as I don't like them. Thanks for the pr though!

